PR #4, opened 2022-07-14
chore(*)move free wrapper into cdefs

chronolaw:
I moved anonymous free functions into `cdefs`,
I think it will make the code more clean and readable.

Does it make sense?

### DIFF
--- a/lib/resty/router/cdefs.lua
+++ b/lib/resty/router/cdefs.lua
@@ -137,4 +137,16 @@ end
 return {
     clib = clib,
     ERR_BUF_MAX_LEN = ERR_BUF_MAX_LEN,
+
+    context_free = function(c)
+        clib.context_free(c)
+    end,
+
+    schema_free = function(s)
+        clib.schema_free(s)
+    end,
+
+    router_free = function(r)
+        clib.router_free(r)
+    end,
 }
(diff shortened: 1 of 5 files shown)


fffonion (approved):

dndx (changes requested):
I have some concerns about this change, basically, we create new closures for each `__gc` metamethod to prevent bugs like https://github.com/Kong/atc-router/pull/3 from happening, after this change, it appears that we might introduce the bug back?

Could you test with the example in https://github.com/Kong/atc-router/pull/3 to see if that is the case? I suspect after this change, it will be the same effect as just using `context = ffi_gc(context, clib.context_free)` directly.

chronolaw:
I copied the code of #3 , then run it with this changes, it print out 'ok'. 

Does it means the code works well? or we can test it in kong code base.

chronolaw:
Also, I add some debug code like `print('s free')` `print('r free')` in cdefs.lua wrapper function, 
When I run test code, the output is :

r free
s free
ok

chronolaw:
If we are not sure about this, I think we can merge it after kong 3.0 release.

fffonion:
`cdefs.context_free` is still a closure, thus i feel it should work as expected. @dndx do we want to pursue with this PR (more tests?) or close it?

dndx (approved):

PR #14, opened 2022-11-16
feat(releases): setup automated releases

hutchic:
implement semmantic release automation such that commit messages that follow conventional commits ( [1](https://github.com/Kong/atc-router/blob/feat/releases/.releaserc#L9) )( [2](https://www.conventionalcommits.org/en/v1.0.0/) ) along with musl / gnu release artifacts for x86_64 and aarch64

image

image

PR #15, opened 2022-11-21
chore(releases): setup continuous delivery and releases

hutchic:
recreation of previous [attempt](https://github.com/Kong/atc-router/pull/14) but using a docker image for toolchain / architecture abstraction purposes.

Commit messages that follow [semantic-commit](https://semantic-release.gitbook.io/semantic-release/) will automatically create a release
image

It will build musl, gnu assets for both amd64 and arm64
image

Using AWS infrastructure via [cirun](https://github.com/Kong/atc-router-cirun/blob/main/.cirun.yml)

Pull requests get a comment annotating the release they are in as well as a `released` label
image

WIP of using the pre-compiled assets [here](https://github.com/Kong/kong-build-tools/pull/607).
Note: build times in some cases reduced from ~44 min musl arm64 to ~25 min
Note: I tried using the rust native toolchain (see previous commits / PR's) it was non-trivial and I gave up
Note: I believe cirun.io requires it's configuration file to exist on the `main` branch before it's able to spin up github actions workers


.cirun.yml
```diff
@@ -0,0 +1,31 @@
+# Self-Hosted Github Action Runners on AWS via Cirun.io
+# Reference: https://docs.cirun.io/Reference/yml.html
+runners:
```
mayocream, line 3: Does this mean the CI will run on Kong's AWS ec2?

hutchic: It does. I tried with docker emulation on github action runner and it doesn't have enough cpu/ram :(

dndx:
Why are we making this so complicated? Rust has perfectly good cross compilation environment that can compile this small project in minutes. Why are we spawning 64-core CPU and 128GB ram `c6g.16xlarge` AWS instance with Docker for the build?

Why do we need to write `Dockerfile` for the build?

See my project: https://github.com/dndx/phantun/blob/main/.github/workflows/release.yml

With mere 50 lines of YAML, it is [releasing into 14 architectures](https://github.com/dndx/phantun/releases/tag/v0.6.0) and the [entire build](https://github.com/dndx/phantun/actions/runs/3322303800) finishes within 7 min using GitHub Actions default x86 runner, and it is a much larger code base than ATC router. Why do we need to introduce `Dockerfile` and run this on a $2.5/hour AWS machine?

This PR makes no sense to me.

hutchic:
@dndx I covered that in the PR message and with some examples as well as not squashing my work which demonstrates I tried the native rust toolchain a few different times.

If the folks more fluent in rust want to take a stab at it I would encourage them to do so

hutchic:
Replaced by https://github.com/Kong/atc-router/pull/16

PR #16, opened 2022-12-01
ci(releases): logic to configure, build, and package releases

hutchic:
Every commit that follows [semantic commit messages](https://www.conventionalcommits.org/en/v1.0.0/) will automatically build (musl, gnu) X (aarch64, x86_64) artifacts

image

Pull requests get automatically commented on and labelled when they're released

image

End result is we don't configure, compile, build, and install atc-router everytime we're building Kong. Instead `curl | tar -C` and it's done

Edit to add: I tested this on my own fork https://github.com/hutchic/atc-router/releases/tag/1.4.9


.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+    name: Create Release Artifacts
+    runs-on: ubuntu-latest
+    container: rustembedded/cross:${{ matrix.target }}
+    strategy:
+      matrix:
+        target:
+          - aarch64-unknown-linux-musl
+          - x86_64-unknown-linux-gnu
+          - x86_64-unknown-linux-musl
+          - aarch64-unknown-linux-gnu
```
dndx, line 42: Do we need `x86_64-unknown-linux-gnu:centos` in the matrix for Glibc 2.17?

hutchic: That would work for x86 but wasn't available for aarch64 so instead https://github.com/Kong/atc-router/pull/16/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R35 covers both options

curiositycasualty: Actually, we might indeed want that. I see both `aarch64-unknown-linux-gnu:centos` and `x86_64-unknown-linux-gnu:centos` available on https://github.com/cross-rs/cross/blob/main/README.md#supported-targets (I must have missed it when I originally looked).

curiositycasualty:
Other than the change to `:centos`, I'd say this is good to go.
I'd feel better about it if @javierguerragiraldez or other rust developers weighed in on it though.

Makefile
```diff
@@ -14,22 +16,22 @@ LUA_INCLUDE_DIR ?= $(PREFIX)/include
...
 INSTALL ?= install
 
+CARGO := $(HOME)/.cargo/bin/cargo
+TARGET ?= x86_64-unknown-linux-gnu
+
 .PHONY: all test install build clean
 
 all: ;
 
 build: target/release/libatc_router.so target/release/libatc_router.a
```
javierguerragiraldez, line 26: i guess this would need to use `$(TARGET)` instead of `release`.   perhaps something like `$(addprefix target/$(TARGET)/, libatc_router.so libatc_router.a)`

.github/workflows/release.yml
dndx: That's why I was asking.

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+    - uses: actions/checkout@v3
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        target: ${{ matrix.target }}
+        override: true
+    - name: Build and Install atc-router
+      env:
+        TARGET: ${{ matrix.target }}
+        RUSTFLAGS: '-C target-feature=-crt-static'
```
dndx, line 53: I'm not sure if we still need this. As `cross` should automatically statically link against `musl` already. We do not generally want to statically link again Glibc.

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+      run: |
+        if [[ '${{matrix.target}}' == 'aarch64-unknown-linux'* ]]; then
+          export RUSTFLAGS="${RUSTFLAGS} -C linker=aarch64-linux-gnu-gcc"
```
dndx, line 56: Could you elaborate on why is this needed?

.github/workflows/release.yml
hutchic: Ran into some challenges handling this with `cross`

```
    strategy:
      matrix:
        target:
          - x86_64-unknown-linux-gnu:centos
#          - aarch64-unknown-linux-gnu:centos
#          - x86_64-unknown-linux-musl
#          - aarch64-unknown-linux-musl

    steps:
      - uses: actions/checkout@v3
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: stable
          target: ${{ matrix.target }}
          override: true
```

Returns
```
nfo: using existing install for 'stable-x86_64-unknown-linux-gnu'
info: override toolchain for '/home/runner/work/atc-router/atc-router' set to 'stable-x86_64-unknown-linux-gnu'

  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.65.0 (897e37553 2022-11-02)

/home/runner/.cargo/bin/rustup target add --toolchain stable x86_64-unknown-linux-gnu:centos
error: toolchain 'stable-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'x86_64-unknown-linux-gnu:centos'
note: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html
Error: The process '/home/runner/.cargo/bin/rustup' failed with exit code 1
```

.github/workflows/release.yml
hutchic: OK I'll try without :+1:

.github/workflows/release.yml
hutchic: https://github.com/cross-rs/cross/blob/main/README.md#supported-targets has me looking at creating a `Cross.toml`. If this is the wrong approach please redirect me :)

.github/workflows/release.yml
hutchic: I tried without and this is necessary for `aarch64-unknown-linux-musl`

dndx:
I am reworking this PR. DO NOT MERGE YET.

colinh-kong:
@dndx anything I can assist with?

dndx:
Closed due to: https://github.com/Kong/atc-router/pull/20#issuecomment-1357307284

PR #18, opened 2022-12-09
fix: handle errors when parsing CStrings

pmalek:
This fixes panics in kong when handling invalid string input data.


dndx:
Superseded by https://github.com/Kong/atc-router/pull/19.

PR #20, opened 2022-12-13
build(bazel) add bazel build and regression test in CI

fffonion:


curiositycasualty:
This PR seems gigantic compared to the github-actions based approach in https://github.com/Kong/atc-router/pull/16
Apart from adding so many autogenerated files to the repo, I have 2 chief concerns:
1. I don't see in this PR any support for compiling for the platforms we care about for the gateway product(s):
  - (older) glibc and musl
  - amd64/x86_64 and arm64/aarch64
2. Today, we don't have to ship the Rust build ecosystem with the product. Unless there's a good reason for packaging `atc-router` as a cargo crate, I don't think adding/shipping the Rust build ecosystem is worth it.

fffonion:
@curiositycasualty this pr doesn't conflict/replace the github actions work that's WIP. it's a seperate track that allows atc-router being built within the new build system.
cross compiling and multiple platforms support will be add in later iterrations.

dndx:
@curiositycasualty I have discussed with @mayocream and @fffonion , they both believe having a binary release does not provide us with much value at the moment with the Bazel based build system. So we are gonna refrain from doing so until it is absolutely needed later.

Just a FYI, this is not some kind of anti-pattern. There are plenty of Rust projects that does not provide binary build, this is especially common with library crates.

mayocream (approved):
LGTM

dndx (approved):

PR #21, opened 2023-01-04
feat(*) support contains operator

pluveto:
FIX KAG-227

### DIFF
--- a/src/semantics.rs
+++ b/src/semantics.rs
@@ -132,10 +132,10 @@ impl Validate for Expression {
                     },
                     BinaryOperator::Contains => {
                         match p.rhs {
-                            Value::String(_) | Value::IpCidr(_) => {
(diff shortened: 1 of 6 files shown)


src/semantics.rs
```diff
@@ -130,6 +130,14 @@ impl Validate for Expression {
                             _ => Err("In/NotIn operators only supports string/IP cidr operands".to_string())
                         }
                     },
+                    BinaryOperator::Contains => {
+                        match p.rhs {
+                            Value::String(_) | Value::IpCidr(_) => {
```
dndx, line 135: We probably shouldn't support `IpCidr` in here as the LHS can only be a single IP address at the point, so `contains` actually makes little sense.

pluveto: fixed

PR #22, opened 2023-01-04
feat(grammar) rawstring

pluveto:
Now we can use \` to construct string so that we don't need to repeatedly write escape slashes.

_Fix KAG-370_

### DIFF
--- a/src/atc_grammar.pest
+++ b/src/atc_grammar.pest
@@ -18,7 +18,7 @@ str_char = { !("\"" | "\\") ~ ANY }
 str_esc = { "\\" ~ ("\"" | "\\" | "n" | "r" | "t") }
 
 rawstr_literal = ${ "`" ~ rawstr_char* ~ "`" }
-rawstr_char = { !"`" ~ ANY}
+rawstr_char = { !"`" ~ ANY }
 
 ipv4_literal = @{ ASCII_DIGIT{1,3} ~ ( "." ~ ASCII_DIGIT{1,3} ){3} }
 ipv6_literal = @{
--- a/src/atc_grammar.pest
+++ b/src/atc_grammar.pest
@@ -17,7 +17,7 @@ str_inner = _{ (str_esc | str_char)* }
 str_char = { !("\"" | "\\") ~ ANY }
 str_esc = { "\\" ~ ("\"" | "\\" | "n" | "r" | "t") }
 
-rawstr_literal = ${"`" ~ rawstr_char* ~ "`"}
+rawstr_literal = ${ "`" ~ rawstr_char* ~ "`" }
 rawstr_char = { !"`" ~ ANY}
 
 ipv4_literal = @{ ASCII_DIGIT{1,3} ~ ( "." ~ ASCII_DIGIT{1,3} ){3} }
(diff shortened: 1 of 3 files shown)


chronolaw:
Could we add one more regex test case for this feature? such as `^/\d+/test$`.

pluveto:
> Could we add one more regex test case for this feature? such as `^/\d+/test$`.

More test cases added

src/atc_grammar.pest
```diff
@@ -17,6 +17,8 @@ str_inner = _{ (str_esc | str_char)* }
 str_char = { !("\"" | "\\") ~ ANY }
 str_esc = { "\\" ~ ("\"" | "\\" | "n" | "r" | "t") }
 
+rawstr_literal = ${"`" ~ rawstr_char* ~ "`"}
```
gszr, line 20:
```suggestion
rawstr_literal = ${ "`" ~ rawstr_char* ~ "`" }
```

src/atc_grammar.pest
```diff
@@ -17,6 +17,8 @@ str_inner = _{ (str_esc | str_char)* }
...
+rawstr_char = { !"`" ~ ANY}
```
gszr, line 21:
```suggestion
rawstr_char = { !"`" ~ ANY }
```

gszr (commented):
Very minor style correction - to align with other rules.

dndx:
Great work, although I have some question regarding the usage of ` for raw strings. Have we considered any other alternatives?

Reason being I have rarely seen ` being used for raw strings in other languages, so am a bit hesitant.

pluveto:
> Great work, although I have some question regarding the usage of ` for raw strings. Have we considered any other alternatives?
> 
> Reason being I have rarely seen ` being used for raw strings in other languages, so am a bit hesitant.

Any suggestion?

gszr:
What about `'` @dndx?

chronolaw:
> What about `'` @dndx?

Perhaps `'` is not a good choice. The user will use the `'` to quote expressions in the command line.

gszr:
> > What about `'` @dndx?
> 
> Perhaps `'` is not a good choice. The user will use the `'` to quote expressions in the command line.

` also has the potential to cause issues in command line usage, as it is reserved for command substitution.

chronolaw:
> how about simply use `_rstart_` `_rend_` as boundary?

In Lua we have `[[ ... ]]`, in C++ we have `R(...)`, we should find an acceptable form.

pluveto:
quotes/parens are common in raw string which means we need to implement escape at least two chars in a raw string literal value.

hbagdi:
@pluveto @gszr @chronolaw Are we still pursuing this one?

chronolaw:
> @pluveto @gszr @chronolaw Are we still pursuing this one?

We still have some discussion on it.

dndx:
@pluveto What about `r""`, just like Python. I think it should not conflict with our grammar rules.

chronolaw:
> @pluveto What about `r""`, just like Python. I think it should not conflict with our grammar rules.

It seems that it is still hard to use `r""` in normal string, like "http.path=r\"...\""

pluveto:
How about rust style:

```rs
let s = r#"'"""''"\/"# // '"""''"\/
```

dndx (approved):

src/parser.rs
```diff
@@ -88,6 +88,18 @@ impl ATCParser {
 
         Ok(s)
     }
+    fn rawstr_literal(input: Node) -> ParseResult<String> {
```
chronolaw, line 91:
```suggestion

    fn rawstr_literal(input: Node) -> ParseResult<String> {
```

PR #23, opened 2023-01-05
fix(*) check all values for equals operator

pluveto:
KAG-228


src/interpreter.rs
```diff
@@ -20,9 +20,10 @@ impl Execute for Expression {
 
                 let (lower, any) = p.lhs.get_transformations();
 
+                // if not in "any" mode, then we need to check all values. `remaining` is the count of unchecked values.
```
chronolaw, line 23: style: break one too-long line into two short lines.

src/interpreter.rs
```diff
@@ -40,7 +42,11 @@ impl Execute for Expression {
                         BinaryOperator::Equals => {
                             if lhs_value == &p.rhs {
                                 m.matches.insert(p.lhs.var_name.clone(), p.rhs.clone());
```
pluveto, line 44: Do we need to record matched value every check time? @dndx

dndx (changes requested):
We might need to do more than just `==`. Every operator should be checked for all values in case LHS is an array.

dndx:
@pluveto need to fix the conflicts

src/interpreter.rs
```diff
@@ -1,222 +1,417 @@
...
+impl Execute for ast::Predicate {
+    fn execute(&self, ctx: &mut Context, m: &mut Match) -> bool {
+        let lhs_values = match ctx.value_of(&self.lhs.var_name) {
+            None => return false,
+            Some(v) => v,
+        };
+
+        let (lower, any) = self.lhs.get_transformations();
+
+        // only can be "all" or "any" mode.
```
dndx, line 30:
```suggestion
        // can only be "all" or "any" mode.
```

src/interpreter.rs
```diff
@@ -1,222 +1,417 @@
...
+        // - all: all values must match (default)
+        // - any: ok if any any matched
+        for mut lhs_value in lhs_values.iter() {
+            let lhs_value_transformed;
```
dndx, line 34:
```suggestion
            let lhs_value_transformed;

```

src/interpreter.rs
```diff
@@ -1,222 +1,417 @@
...
+            if lower {
+                match lhs_value {
+                    Value::String(s) => {
+                        lhs_value_transformed = Value::String(s.to_lowercase());
+                        lhs_value = &lhs_value_transformed;
+                    }
+                    _ => unreachable!(),
+                }
+            }
```
dndx, line 43:
```suggestion
            }

```

src/interpreter.rs
```diff
@@ -1,222 +1,417 @@
...
+            let mut curr_match = false;
```
dndx, line 44:
```suggestion
            let mut matched = false;
```

`curr_match` was a bit hard to catch. I had to keep reading to understand what it does.

src/interpreter.rs
```diff
@@ -1,222 +1,417 @@
...
+            match self.op {
+                BinaryOperator::Equals => {
+                    if lhs_value == &self.rhs {
+                        m.matches
+                            .insert(self.lhs.var_name.clone(), self.rhs.clone());
+                        curr_match = true;
+                        if any {
+                            return true;
+                        }
```
dndx, line 53: This check can be moved to before the above line, as if this is `true` then above line does not needs execution.

src/interpreter.rs
```diff
@@ -1,222 +1,417 @@
...
+                    };
 
-                // no match for all values
-                false
+                    if lhs.contains(rhs) {
+                        return true;
+                    }
+                }
+            } // match
+            if !any && !curr_match {
```
dndx, line 257:
```suggestion

            if !any && !curr_match {
```

Style: add a newline before and after `if` blocks.

dndx (changes requested):

src/interpreter.rs
```diff
@@ -1,222 +1,432 @@
...
-                                lhs_value_transformed = Value::String(s.to_lowercase());
-                                lhs_value = &lhs_value_transformed;
-                            }
-                            _ => unreachable!(),
+            Expression::Predicate(p) => p.execute(ctx, m),
+        }
+    }
+}
+
+impl Execute for ast::Predicate {
```
dndx, line 21:
```suggestion
impl Execute for Predicate {
```

Let's import `Predicate` directly to stay consistent with the style of this project.

src/interpreter.rs
```diff
@@ -1,222 +1,432 @@
...
 
-                // no match for all values
-                false
+                    if lhs.contains(rhs) {
+                        return true;
+                    }
+                }
+            } // match
+
+            if !any && !matched {
```
dndx, line 272:
```suggestion
            if !any && !matched {
                // all and nothing matched
```

src/interpreter.rs
```diff
@@ -1,222 +1,432 @@
...
+                return false;
             }
-        }
+        } // for iter
+
+        return lhs_values.len() > 0 && !any;
```
dndx, line 277:
```suggestion
        // if we reached here, it means that `any` did not find a match,
        // or we passed all matches for `all`. So we simply need to return
        // !any && lhs_values.len() > 0 to cover both cases
        return !any && lhs_values.len() > 0;
```

src/interpreter.rs
```diff
@@ -1,222 +1,436 @@
...
+fn test_predicate() {
+    let mut mat = Match::new();
+    let mut schema = schema::Schema::default();
+    schema.add_field("my_key", Type::String);
+    let mut ctx = Context::new(&schema);
+
+    // check when value list is empty
+    // check if all values match starts_with foo -- should be false
+    let p = Predicate {
+        lhs: ast::Lhs {
```
dndx, line 295: I noticed that the `ast::` namespace is only used in tests. In this case we can write `use crate::ast;` inside `test_predicate` instead since the above code does not use it.

dndx (approved):

PR #27, opened 2023-01-18
KAG-458: ATC router limits Route.paths to 2048 bytes

hanshuebner:


src/ffi.rs
```diff
@@ -177,8 +179,9 @@ pub unsafe extern "C" fn context_add_value(
 
     let value: Result<Value, _> = value.try_into();
     if let Err(e) = value {
-        errbuf[..e.len()].copy_from_slice(e.as_bytes());
-        *errbuf_len = e.len();
+        let errlen = min(e.len(), ERR_BUF_MAX_LEN);
```
flrgh, line 182: We will probably want a similar test case for `context:add_value()` to cover this.

dndx: I double checked and the only error that could happen with `add_value` is `Utf8Error`, which can not possibly exceed `2048` bytes due to the formatting:

https://doc.rust-lang.org/src/core/str/error.rs.html#115-122

So maybe we only need to handle the parsing error. You should not even be able write a test case for `add_value` that triggers this bug.

src/parser.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(deprecated)]
```
kikito, line 1: Unless the warnings are preventing us from doing something with the library, I would keep them on.

kikito (approved):
In my limited Rust understanding, this PR looks ok.
* It duplicates the buffer size
* It truncates error messages so that they are not longer than the buffer size in two places
* It adds unit tests for faster iteration
* It also has an "integration" test that duplicates the problem from the Nginx/Lua side
I have a minor comment about deprecation warnings but other than that LGTM

src/parser.rs
hanshuebner: > Unless the warnings are preventing us from doing something with the library, I would keep them on.

We are using deprecated features of some underlying library, but as this is a known issue _and_ we're going to see things fail if we upgrade these libraries, switching the deprecation messages off is better because it makes feature work and fixes much easier because one does not have to look through pages of useless warnings to find the relevant errors.

dndx: IMO these warnings should be kept. We are already in the process of migrating off `prec_climber` and the deprecation warnings should not be causing tests and clippy checks to fail.

PR #28, opened 2023-01-26
chore(build): remove cargo binary check from Makefile

flrgh:
This check has proven to be somewhat error-prone, and it's a little bit handhold-y for a Makefile. If cargo is truly not installed, this error seems plenty explanatory for an average developer:

```
$ make build
cargo build --release
make: cargo: No such file or directory
make: *** [Makefile:27: target/release/libatc_router.so] Error 127
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,6 @@ all: ;
 build: target/release/libatc_router.so target/release/libatc_router.a
 
 target/release/libatc_router.%: src/*.rs
-ifeq (, $(shell cargo))
-$(error "cargo not found in PATH, consider doing \"curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh\"")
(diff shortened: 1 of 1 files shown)


flrgh:
found a fix for this in bazel

fffonion:
@flrgh Thanks for making the fix. I agree the check might be not necessary, and because of the way Makefile is evaluated, the command is ran at load time, meaning we are not planning to build the .so, the sanity check fails, which is not strictlly right.

Maybe let's just remove the check? I do agree the error is self-explantory.

flrgh:
> Maybe let's just remove the check? I do agree the error is self-explantory.

@fffonion works for me; updated the PR for this

fffonion (approved):

fffonion:
Thank you @flrgh !

PR #46, opened 2023-03-08
chore(ci): experimental lualint integration (use github action)

pluveto:


pluveto:
We have implemented a line-level filter that will only report errors on lines that have been changed.

PR #82, opened 2023-07-04
fix(semantics): fix incorrect type checking that prevented `IpAddr in

dndx:
IpCidr` matchings

https://konghq.atlassian.net/browse/KAG-1846

### DIFF
--- a/src/semantics.rs
+++ b/src/semantics.rs
@@ -81,7 +81,11 @@ impl Validate for Expression {
                 }
                 let lhs_type = lhs_type.unwrap();
 
-                if p.op != BinaryOperator::Regex && lhs_type != &p.rhs.my_type() {
+                if p.op != BinaryOperator::Regex // Regex RHS is always Regex, and LHS is always String
+                    && p.op != BinaryOperator::In // In/NotIn supports String in String or IPAddr in IpCidr
+                    && p.op != BinaryOperator::NotIn
+                    && lhs_type != &p.rhs.my_type()
+                {
                     return Err(
                         "Type mismatch between the LHS and RHS values of predicate".to_string()
                     );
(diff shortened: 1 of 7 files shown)


src/semantics.rs
```diff
@@ -123,11 +127,11 @@ impl Validate for Expression {
                         }
                     },
                     BinaryOperator::In | BinaryOperator::NotIn => {
-                        match p.rhs {
-                            Value::String(_) | Value::IpCidr(_) => {
+                        match (lhs_type, &p.rhs,) {
```
chronolaw, line 130: Should be `(lhs_type, &p.rhs)`? (remove `,`)?

dndx: This is by auto format, and is the recommended format for tuples by Rustfmt.

t/07-in_notin.t
```diff
@@ -0,0 +1,142 @@
...
+nilIn/NotIn operators only supports string in string or IP in CIDR
+nilIn/NotIn operators only supports string in string or IP in CIDR
+--- no_error_log
+[error]
+[warn]
+[crit]
+
+
+
+=== TEST 2: in operator works with String operands
```
chronolaw, line 59: We should add `notin` test case here.

t/07-in_notin.t
```diff
@@ -0,0 +1,142 @@
...
+            local schema = require("resty.router.schema")
+            local router = require("resty.router.router")
+            local context = require("resty.router.context")
+
+            local s = schema.new()
+
+            s:add_field("l3.ip", "IpAddr")
+
+            local r = router.new(s)
+            assert(r:add_matcher(0, "a921a9aa-ec0e-4cf3-a6cc-1aa5583d150c",
```
chronolaw, line 119: We should add `notin` test case here.

dndx: Added

chronolaw (approved):

PR #105, opened 2023-08-22
feat(serde): add serde integration

sumimakito:
This pull request is a prerequisite for the external crate containing ATC Router's WASM binding

- Adding `lib` to `crate-type` to allow this crate to be used by other crates
- Adding a `serde` feature to apply `serde` macros selectively
- Adding a `no-ffi` feature to exclude `ffi` from mods to avoid symbol conflicts while being used as a dependency

KAG-2426

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ lazy_static = "1.4"
 uuid = "1.5"
 regex = "1.10"
 serde = { version = "1.0", features = ["derive"], optional = true }
-serde_regex = { version = "1.1.0", optional = true }
+serde_regex = { version = "1.1", optional = true }
 
 [lib]
 crate-type = ["lib", "cdylib", "staticlib"]
@@ -25,4 +25,4 @@ crate-type = ["lib", "cdylib", "staticlib"]
 [features]
 default = ["ffi"]
 ffi = []
-serde = ["cidr/serde", "dep:serde", "serde_regex"]
+serde = ["cidr/serde", "dep:serde", "dep:serde_regex"]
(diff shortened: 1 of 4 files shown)


dndx (commented):
https://rust-lang.github.io/rust-clippy/master/index.html#negative_feature_names

Clippy seems to think `no-ffi` is a poor name for feature. Instead, it suggests to use:

```
[features]
default = ["ffi"]
```

instead.

sumimakito:
@dndx I have updated the branch with the suggestion. Thanks for pointing it out - I was not aware of the default features before.

dndx:
@sumimakito Could you rebase this?

Cargo.toml
```diff
@@ -16,6 +16,13 @@ cidr = "0.2"
...
+serde_regex = { version = "1.1.0", optional = true }
 
 [lib]
-crate-type = ["cdylib", "staticlib"]
+crate-type = ["lib", "cdylib", "staticlib"]
+
+[features]
+default = ["ffi"]
+ffi = []
+serde = ["cidr/serde", "dep:serde", "serde_regex"]
```
dndx, line 28:
```suggestion
serde = ["cidr/serde", "dep:serde", "dep:serde_regex"]
```

? Is `serde_regex` expected?

Cargo.toml
```diff
@@ -16,6 +16,13 @@ cidr = "0.2"
 lazy_static = "1.4"
 uuid = "1.5"
 regex = "1.10"
+serde = { version = "1.0", features = ["derive"], optional = true }
+serde_regex = { version = "1.1.0", optional = true }
```
dndx, line 20:
```suggestion
serde_regex = { version = "1.1", optional = true }
```

To allow dependabot to bump dependency versions.

Cargo.toml
sumimakito: No, I missed that out. I added it back.

dndx (approved):

PR #175, opened 2024-01-18
refactor(lib): validate() will return fields

chronolaw:
We will use the fields to do more check.

### DIFF
--- a/lib/resty/router/router.lua
+++ b/lib/resty/router/router.lua
@@ -122,7 +122,7 @@ do
 
         assert(r:remove_matcher(DEFAULT_UUID))
 
-        return true
+        return r:get_fields()
     end
 end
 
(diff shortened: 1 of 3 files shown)


lib/resty/router/router.lua
```diff
@@ -120,9 +120,11 @@ do
             return nil, err
         end
 
+        local fields = r:get_fields()
```
dndx, line 123: Error handling is missing.

dndx (commented):
Need to update `README` as well.

lib/resty/router/router.lua
chronolaw: doc changed.

dndx (approved):

ADD-SP (approved):

PR #181, opened 2024-01-22
chore(tests): fix broken OpenSSL build

dndx:
Older releases are no longer available, bump to their respective latest patch version.


.github/workflows/tests.yml
```diff
@@ -13,29 +13,35 @@ concurrency:
...
+        - nginx: "1.25.3"
+          nginx_opts: "--without-pcre2"
+          openssl: "3.2.0"
+          valgrind: "valgrind"
+          lua_nginx_module: "v0.10.26"
+          lua_resty_core: "v0.1.28"
         - nginx: "1.23.0"
           nginx_opts: "--without-pcre2"
-          openssl: "3.1.1"
+          openssl: "3.1.4"
```
AndyZhang0707, line 29: Don't need to fix this time, but we need to think how can we handle the situation when this version becomes an "old" version or are we going to keep using the latest version.

AndyZhang0707 (approved):

AndyZhang0707 (approved):

chobits (approved):

.github/workflows/tests.yml
dndx: Added comment

PR #183, opened 2024-01-24
feat(lib): add "not" logical operator support

dndx:
The not operator `!` can be used with parenthesized expressions and will negate the result of the expressions.

Pratt parser is not used for this prefix operator because that would make:

```
! a == 1
```

Possible, which may cause confusions. Instead, this was implemented on the parenthesized expressions rule only so you can only write `! (a == 1)`.

[KAG-3605]

[KAG-3605]: https://konghq.atlassian.net/browse/KAG-3605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -18,6 +18,7 @@ pub enum Expression {
 pub enum LogicalExpression {
     And(Expression, Expression),
     Or(Expression, Expression),
+    Not(Expression),
 }
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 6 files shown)


chronolaw (approved):

sumimakito (approved):

dndx:
Which one do you like better? `! (a == 1)` or `not (a == 1)`?

sumimakito:
> Which one do you like better? `! (a == 1)` or `not (a == 1)`?

@dndx As we now have `||` and `&&` instead of `or` and `and` in our grammar, IMO `!` looks more consistent than `not`.

chobits (approved):

chobits:
+ 1 for `!` which is more concise

PR #244, opened 2024-08-04
perf(lib): add `#[inline]` for small functions

fjchen7:
`#[inline]` on function can tell compiler to automatically expand the function at the places where it's used. The cost is that the the compiled binary file will become larger, and the compilation time might increase. But in general experience the trade-off are usually worth it for small function.

Below is the benchmark result of a hot spot function `Predicate::Execute` method, with efficiency improved by nearly 6.7%, in macOS on Apple M1 Pro (10-core, 32GB RAM).
```
time:   [6.4486 ns 6.4725 ns 6.5054 ns]
change: [-12.594% -6.7029% -2.1903%] (p = 0.01 < 0.05)
Performance has improved.
```

The benchmark source code

```rust
fn criterion_benchmark(c: &mut Criterion) {
    let mut schema = schema::Schema::default();
    schema.add_field("my_key", ast::Type::String);

    let p = Predicate {
        lhs: ast::Lhs {
            var_name: "my_key".to_string(),
            transformations: vec![],
        },
        rhs: Value::String("foo".to_string()),
        op: BinaryOperator::Prefix,
    };

    c.bench_function("Predicate::execute", |b| {
        b.iter(|| {
            let mut ctx = Context::new(&schema);
            let mut mat = Match::new();
            p.execute(&mut ctx, &mut mat)
        })
    });
}
```

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -51,6 +51,7 @@ macro_rules! parse_num {
 }
 
 impl ATCParser {
+    #[inline]
     fn new() -> Self {
         Self {
             pratt_parser: PrattParser::new()
@@ -70,6 +71,7 @@ impl ATCParser {
     }
 }
 
+#[inline]
 fn parse_ident(pair: Pair<Rule>) -> ParseResult<String> {
     Ok(pair.as_str().into())
 }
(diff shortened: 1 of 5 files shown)


src/parser.rs
```diff
@@ -51,6 +51,7 @@ macro_rules! parse_num {
 }
 
 impl ATCParser {
+    #[inline]
```
dndx, line 54: This might be a bit too much, `ATCParser` is not on a hot codepath and I'm not sure if inlining even helps here.

dndx (changes requested):

src/parser.rs
fjchen7: If the code path isn't hot, the benefit might be not much significant. I've reverted it.

PR #246, opened 2024-09-04
perf(interpreter): improve ATC interprater's cache efficiency

xianghai2:
implementation for [KAG-5152](https://konghq.atlassian.net/browse/KAG-5152)

[KAG-5152]: https://konghq.atlassian.net/browse/KAG-5152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

What is done with this PR:
1. introduce CirInstruction intermediate representation
2. replace 'Expression' of 'matchers' with 'CirProgram'

Why needs such a change:
Original 'Expression' is implemented as nested via Box, this leads to
separated multiple heap memory regions for an individual matcher, and
it costs more memory access time and memory overhead.


src/ast.rs
```diff
@@ -6,6 +6,34 @@ use std::net::IpAddr;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub struct Route {
```
ADD-SP, line 11: Do you think we should use a more specific name? `Router`, `Route`, `RouteTerm`, and `RouteLogicalOperators`, which looks weird.

src/interpreter.rs
```diff
@@ -18,6 +78,68 @@ impl Execute for Expression {
...
+                        }
+                    }
+                }
+                RouteTerm::Predicate(p) => {
+
+                    arr_predicate.push(OperandItem::Predicate(p));
+                }
+            }
+        }
+        return evaluate_operand_item(arr_predicate.pop().unwrap(), ctx, m);
```
StarlightIbuki, line 139: Should we check if this is the last item in the stack?

src/router.rs
```diff
@@ -12,49 +14,53 @@ struct MatcherKey(usize, Uuid);
...
 
-        let ast = parse(atc).map_err(|e| e.to_string())?;
-
-        ast.validate(self.schema)?;
-        ast.add_to_counter(&mut self.fields);
-
-        assert!(self.matchers.insert(key, ast).is_none());
+        let route = self.convert_route(parse(atc).map_err(|e| e.to_string())?);
+        route.validate(self.schema)?;
+        route.add_to_counter(&mut self.fields);
```
StarlightIbuki, line 45: Let's validate the AST first and then convert it to IR. IR strips out extra info which is necessary for the validator to produce user-friendly error messages.

src/semantics.rs
```diff
@@ -159,6 +159,142 @@ impl Validate for Expression {
...
+                    if *val == 0 {
+                        assert!(map.remove(&p.lhs.var_name).is_some());
+                    }
+                }
+            }
+        }
+    }
+}
+
+impl Validate for Route {
```
StarlightIbuki, line 195: We should do this on AST instead of on Route.

StarlightIbuki (approved):
Would it be beneficial if we use Vec instead of Hashmap for storing routers?

StarlightIbuki (approved):
Would it be beneficial if we use Vec instead of Hashmap for storing routers?

StarlightIbuki (approved):
Would it be beneficial if we use Vec instead of Hashmap for storing routers?

src/ast.rs
```diff
@@ -6,6 +6,34 @@ use std::net::IpAddr;
...
+    pub stack: Vec<RouteTerm>,
+}
+
+impl Route {
+    #[must_use]
```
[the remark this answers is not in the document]
xianghai2: Thanks. 
Removed it. Original intention was a reminder.

src/interpreter.rs
```diff
@@ -1,10 +1,70 @@
...
+                }
+            },
+            Expression::Predicate(p) => {
+                let mut predicate = Predicate {
+                    lhs: crate::ast::Lhs {
+                        var_name: p.lhs.var_name.clone(),
+                        transformations: Vec::new(),
+                    },
+                    rhs: p.rhs.clone(),
+                    op: p.op.clone(),
```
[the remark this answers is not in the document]
xianghai2: Fixed, use copy.

src/interpreter.rs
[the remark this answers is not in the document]
xianghai2: I would keep this in case it's needed for further use cases of 'Expression' in future.

src/ast.rs
xianghai2: The name of 'Route' is derived from the existing 'Router' container. The meaning is that Router container is a collection of individual Route.  And the RouteTerm is a subcomponent of a Route item.

src/interpreter.rs
```diff
@@ -1,10 +1,70 @@
...
 pub trait Execute {
     fn execute(&self, ctx: &mut Context, m: &mut Match) -> bool;
 }
 
+pub trait Convert {
+    fn convert(&self, route: &mut Route) -> bool;
+}
+
+impl Convert for Expression {
+    fn convert(&self, route: &mut Route) -> bool {
```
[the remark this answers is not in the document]
xianghai2: Fixed to return nothing, thanks.

ADD-SP (commented):
There are two IRs in this implementation, one is AST, another one is StackOps. Could you split its codes into two different modules?

src/router.rs
xianghai2: Fixed, validate the AST first. 
And here, at the moment, Route does not strip any info. Just in case, Expression will add more information in future.

src/semantics.rs
xianghai2: Route also need validation, in case there may be anything wrong with converion.

src/interpreter.rs
```diff
@@ -1,10 +1,63 @@
...
+                    route
+                        .stack
+                        .push(RouteTerm::LogicalOperator(RouteLogicalOperators::Not));
+                }
+            },
+            Expression::Predicate(p) => {
+                let mut predicate = Predicate {
+                    lhs: crate::ast::Lhs {
+                        var_name: p.lhs.var_name.clone(),
+                        transformations: Vec::new(),
```
ProBrian, line 44: Could we just init this like `transformations: p.lhs.transformations.clone(),`, so that we don't need the iteration of transformation assignment below.

src/interpreter.rs
```diff
@@ -1,10 +1,70 @@
-use crate::ast::{BinaryOperator, Expression, LogicalExpression, Predicate, Value};
+use crate::ast::{
+    BinaryOperator, Expression, LogicalExpression, Predicate, Route, RouteLogicalOperators,
+    RouteTerm, Value,
+};
 use crate::context::{Context, Match};
-
+use stack_array::*;
```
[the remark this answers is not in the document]
xianghai2: Removed the use of that external library.

src/interpreter.rs
xianghai2: not really.

src/interpreter.rs
```diff
@@ -1,10 +1,70 @@
...
 pub trait Execute {
     fn execute(&self, ctx: &mut Context, m: &mut Match) -> bool;
 }
 
+pub trait Convert {
+    fn convert(&self, route: &mut Route) -> bool;
+}
```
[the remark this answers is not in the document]
xianghai2: Fixed, return nothing. thanks.

src/ast.rs
```diff
@@ -6,6 +6,34 @@ use std::net::IpAddr;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub struct Route {
+    pub stack: Vec<RouteTerm>,
+}
+
+impl Route {
```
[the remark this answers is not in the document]
xianghai2: Fixed, moved to a separate module/file.  thanks.

src/interpreter.rs
xianghai2: Fixed. thanks.

src/line.rs
```diff
@@ -0,0 +1,30 @@
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub struct Route {
+    pub stack: Vec<RouteTerm>,
+}
+
+impl Route {
+    pub fn new() -> Self {
+        Self { stack: Vec::new() }
+    }
```
StarlightIbuki: Please address the lint issue

src/interpreter.rs
```diff
@@ -18,6 +64,99 @@ impl Execute for Expression {
...
+        let mut operand_stack: [OperandItem; 2] = [OperandItem::Val(false); 2];
+
+        for item in &self.stack {
+            match item {
+                RouteTerm::LogicalOperator(op) => {
+                    match op {
+                        RouteLogicalOperators::And => {
+                            //stack pop
+                            top -= 1;
+                            let left = evaluate_operand_item(operand_stack[top], ctx, m);
```
ProBrian, line 92: Seems the we use postorder iteration to store expressions in `self.stack` so the stack would like: `[left, right, op, left, right, op]`. and the operand stack will also be [left, right] when we iterating `self.stack` here. so when we "pop" the operand stack, seems we will first get "right" then "left"? That will not affect the correctness, but maybe make accurate naming will make it easy to understand.

xianghai2: Correct. Fixed. thanks.

src/interpreter.rs
```diff
@@ -18,6 +64,95 @@ impl Execute for Expression {
...
+impl Execute for Route {
+    fn execute(&self, ctx: &mut Context, m: &mut Match) -> bool {
+        //operand stack
+        let mut top: usize = 0;
+        let mut operand_stack: [OperandItem; 2] = [OperandItem::Val(false); 2];
+
+        for item in &self.stack {
+            match item {
+                RouteTerm::LogicalOperator(op) => {
+                    match op {
```
StarlightIbuki, line 88: It's possible to match with `RouteTerm::LogicalOperator(RouteLogicalOperators::And)` Thus we do not need to nest 2 levels of match.

src/interpreter.rs
```diff
@@ -18,6 +64,95 @@ impl Execute for Expression {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum OperandItem<'a> {
+    Val(bool),
+    Predicate(&'a Predicate),
```
StarlightIbuki, line 70: I doubt whether it is worthwhile to lazy-eval pred. Pred is quite simple to calculate but copying a pred involves copying strings. Let's examine the actual influence.
IMO lazy-eval works if we go top-down, but it does not make a lot of difference if we always calculate every logic operator.

StarlightIbuki: After rethinking, it's only making a reference thus it's acceptable, and the short-circuit always happens at the left-hand side thus it's effectively the same as top-down short-circuit.

src/ast.rs
```diff
@@ -6,6 +6,34 @@ use std::net::IpAddr;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub struct Route {
+    pub stack: Vec<RouteTerm>,
```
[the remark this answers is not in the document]
xianghai2: fixed

src/ast.rs
xianghai2: fixed

src/linear.rs
ADD-SP, line 1: Should be named `lir.rs`.

src/interpreter.rs
```diff
@@ -18,6 +19,95 @@ impl Execute for Expression {
...
+                            if left {
+                                // short circuit
+                                // stack pop
+                                top -= 2;
+
+                                //stack push
+                                operand_stack[top] = OperandItem::Val(true);
+                                top += 1;
+                            } else {
+                                let right = evaluate_operand_item(operand_stack[top - 1], ctx, m);
```
StarlightIbuki, line 75: Actually, it can be simplified as:

```rust
operand_stack[0] = evaluate_operand_item(operand_stack[top - 2], ctx, m) or evaluate_operand_item(operand_stack[top - 1], ctx, m);
top = 1;
```

Which is more intuitive and straightforward, and also shorter.

src/router.rs
```diff
@@ -12,49 +14,55 @@ struct MatcherKey(usize, Uuid);
...
         }
 
         let ast = parse(atc).map_err(|e| e.to_string())?;
-
         ast.validate(self.schema)?;
-        ast.add_to_counter(&mut self.fields);
-
-        assert!(self.matchers.insert(key, ast).is_none());
+        let lir = self.translate_lir(ast);
+        lir.validate(self.schema)?;
```
StarlightIbuki, line 46: This step is unnecessary.

StarlightIbuki (commented):
Please rebase to solve the lint issue

src/linear.rs
xianghai2: fixed, thanks.

StarlightIbuki (approved):
LGTM

src/interpreter.rs
Oyami-Srk: This also could apply to the And Operator. Using the short-circuit provided by language first will be more straightforward.

src/interpreter.rs
```diff
@@ -18,6 +19,95 @@ impl Execute for Expression {
...
+                    // push stack
+                    operand_stack[top] = OperandItem::Predicate(p);
+                    top += 1;
+                }
+            }
+        }
+        //stack pop
+        top -= 1;
+        if top != 0 {
+            panic!("There may be something wrong for the AST conversion!")
```
Oyami-Srk: Hint of validation: https://github.com/Kong/atc-router/blob/3f8f648b344d783e076b0f197cdd93102bd3d535/src/semantics.rs#L65

Oyami-Srk (approved):

ADD-SP (approved):

src/semantics.rs
```diff
@@ -13,55 +14,6 @@ pub trait FieldCounter {
     fn remove_from_counter(&self, map: &mut HashMap<String, usize>);
 }
 
-impl FieldCounter for Expression {
```
dndx, line 16: Why not keep `FieldCounter` operating on the AST, not `Lir` for a few reasons:

1. It allows us flexibility for changing the VM implementation
2. It clearly separating the responsibility of VM and Parser, which results structurally better code
3. Semantics does not depends on `Lir` implementation details

Could you explain if there are anything that I'm missing that might result in different decision being made?

dndx (changes requested):

src/semantics.rs
ADD-SP: In my opinion, the `FieldCounter` is not a feature of a DSL spec, it is a feature of its usage (atc-router), so we should not implement it on AST as the responsibility of AST is to reflect the spec of the DSL

The `atc-router` must have the ability to do incremental rebuild, so we need the `FieldCounter`, this is the requirement of the `atc-router`, which is not in the spec of `atc-lang` yet.

The `Lir` is a part of the implementation for `atc-lang`, which is used by the `atc-router`. For the best practice of type-level programming, the `FieldCounter` should be implemented by `atc-router`, or `atc-vm` that is used by the `atc-router`.

Currently, we haven't split the `atc-router` into multiple crates, I think we should do it the make the responsibility and dataflow more clear.

ADD-SP: Deriving further.

For the use case that needs incremental rebuild, we should choose a VM that supports incremental rebuild and sacrifices the opportunities of optimization.

For the use case that doesn't need it, we should also choose a VM that can do more optimization by sacrificing incremental build

ADD-SP: Instead of type-level programming, from the implementation perspective, we have lost the old ASTs when adding/removing a matcher, so keeping the `FieldCounter` for AST doesn't help to rebuild, so we removed the dead code yo reduce the maintenance burden.

src/lir.rs
```diff
@@ -0,0 +1,145 @@
...
+                    .push(LirInstruction::LogicalOperator(LirLogicalOperators::Or));
+            }
+            LogicalExpression::Not(r) => {
+                translate_helper(r, lir);
+                lir.program
+                    .push(LirInstruction::LogicalOperator(LirLogicalOperators::Not));
+            }
+        },
+        Expression::Predicate(p) => {
+            let predicate = Predicate {
```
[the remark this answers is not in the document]
StarlightIbuki: > Not really, as Predicate and Lhs does not implement clone(). thanks.

what about implement Clone

src/interpreter.rs
```diff
@@ -1,20 +1,75 @@
...
-        match self {
-            Expression::Logical(l) => match l.as_ref() {
-                LogicalExpression::And(l, r) => l.execute(ctx, m) && r.execute(ctx, m),
-                LogicalExpression::Or(l, r) => l.execute(ctx, m) || r.execute(ctx, m),
-                LogicalExpression::Not(r) => !r.execute(ctx, m),
-            },
-            Expression::Predicate(p) => p.execute(ctx, m),
+        //operand stack
+        let mut top: usize = 0;
+        let mut operand_stack: [OperandItem; 2] = [OperandItem::Val(false); 2];
```
StarlightIbuki, line 27: This does not seem a sound implementation. Have you tried example like

```
(a and b) or (c and d)
```

StarlightIbuki (changes requested):
Dismiss formal approval

src/interpreter.rs
xianghai2: fixed, fall back to initial implementation. thanks.

src/lir.rs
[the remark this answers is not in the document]
xianghai2: Fixed, add Clone for Predicate and Lhs. thanks.

src/interpreter.rs
```diff
@@ -1,20 +1,53 @@
...
+impl Execute for LirProgram {
     fn execute(&self, ctx: &mut Context, m: &mut Match) -> bool {
-        match self {
-            Expression::Logical(l) => match l.as_ref() {
-                LogicalExpression::And(l, r) => l.execute(ctx, m) && r.execute(ctx, m),
-                LogicalExpression::Or(l, r) => l.execute(ctx, m) || r.execute(ctx, m),
-                LogicalExpression::Not(r) => !r.execute(ctx, m),
-            },
-            Expression::Predicate(p) => p.execute(ctx, m),
+        let mut operand_stack: Vec<OperandItem> = Vec::new();
```
StarlightIbuki, line 25: IMO this does not perform better than recursion implementation, plus it cannot short-circuit the right branch of an `and` or `or`.

xianghai2:
Convert to draft and need more testing and re-work. thanks.

src/semantics.rs
dndx: @ADD-SP For our use case, sacrefising incremental rebuild is basically a no-go. But having field information at parse time is not only critical for Kong, but also for a lot of projects that use this crate. I recommend us to consider more carefully before reaching a conclusion.

src/lir.rs
```diff
@@ -0,0 +1,145 @@
+use crate::ast::{Expression, LogicalExpression, Predicate};
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub struct Lir {
+    pub program: Vec<LirInstruction>,
```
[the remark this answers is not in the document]
dndx: Then the design is probably not correct, at most it should be `pub(crate)`, fully public in the wild makes little sense.

src/cir.rs
```diff
@@ -0,0 +1,439 @@
...
+fn execute_helper(
+    cir_instructions: &[CirInstruction],
+    index: usize,
+    ctx: &mut Context,
+    m: &mut Match,
+) -> bool {
+    match &cir_instructions[index] {
+        CirInstruction::AndIns(and) => {
+            let left_val = if is_index(&and.left) {
+                execute_helper(cir_instructions, and.left.as_index().unwrap(), ctx, m)
```
ProBrian, line 259: Does The deepest recursion depth here as same as the previous `Expression::execution` if no short circuit happens here?

src/cir.rs
```diff
@@ -0,0 +1,435 @@
...
+            };
+
+            if !left_val {
+                // short circuit
+                false
+            } else if is_index(&and.right) {
+                execute_helper(cir_instructions, and.right.as_index().unwrap(), ctx, m)
+            } else {
+                and.right.as_predicate().unwrap().execute(ctx, m)
+            }
```
ProBrian, line 273:
```suggestion
            }
```

src/cir.rs
```diff
@@ -0,0 +1,439 @@
...
+            if !left_val {
+                // short circuit
+                return false;
+            } else {
+                if is_index(&and.right) {
+                    return execute_helper(cir_instructions, and.right.as_index().unwrap(), ctx, m);
+                } else {
+                    return and.right.as_predicate().unwrap().execute(ctx, m);
+                }
+            }
```
ProBrian, line 273: left_val && if is_index(&and.right) {
                execute_helper(cir_instructions, and.right.as_index().unwrap(), ctx, m)
            } else {
                and.right.as_predicate().unwrap().execute(ctx, m)
            }
```suggestion

            if !left_val {
                // short circuit
                return false;
            } else {
                if is_index(&and.right) {
                    return execute_helper(cir_instructions, and.right.as_index().unwrap(), ctx, m);
                } else {
                    return and.right.as_predicate().unwrap().execute(ctx, m);
                }
            }
```

ProBrian, line 273:
```suggestion
            left_val && if is_index(&and.right) {
                execute_helper(cir_instructions, and.right.as_index().unwrap(), ctx, m)
            } else {
                and.right.as_predicate().unwrap().execute(ctx, m)
            }
```
What about using a short circuit pattern like this

src/cir.rs
```diff
@@ -0,0 +1,439 @@
...
+) {
+    loop {
+        if operator_stack.len() > 0 {
+            match &operator_stack.last().unwrap() {
+                LirLogicalOperators::And => {
+                    if operand_stack.len() >= 2 {
+                        let right = operand_stack.pop().unwrap();
+                        let left = operand_stack.pop().unwrap();
+                        let and_ins = AndIns {
+                            left: left.clone(),
```
ProBrian, line 103: Seems that we don't need `clone` here, since `left` and `right` are `pop`ped from operand, we can just move their ownership

src/cir.rs
```diff
@@ -0,0 +1,439 @@
...
+                    } else {
+                        break;
+                    }
+                }
+                LirLogicalOperators::Or => {
+                    if operand_stack.len() >= 2 {
+                        let right = operand_stack.pop().unwrap();
+                        let left = operand_stack.pop().unwrap();
+                        let or_ins = OrIns {
+                            left: left.clone(),
```
ProBrian, line 118: same here, no need to `clone`

src/cir.rs
```diff
@@ -0,0 +1,439 @@
...
+        execute_helper(&self.instructions, self.instructions.len() - 1, ctx, m)
+    }
+}
+
+fn add_to_counter_helper(
+    cir_instructions: &[CirInstruction],
+    index: usize,
+    map: &mut HashMap<String, usize>,
+) {
+    match &cir_instructions[index] {
```
ProBrian, line 315: Is that correct that we don't do iteration for `cir_instructions` here? It seems like that if we have a `AndIns(index, index)` in the tail of `cir_instructions`, than nothing will be done for the fields counter?

src/cir.rs
```diff
@@ -0,0 +1,439 @@
...
+        }
+    }
+}
+
+fn remove_from_counter_helper(
+    cir_instructions: &[CirInstruction],
+    index: usize,
+    map: &mut HashMap<String, usize>,
+) {
+    match &cir_instructions[index] {
```
ProBrian, line 350: Same question here, we should iterator the `cir_instructions` to reduce counter for every field

src/cir.rs
```diff
@@ -0,0 +1,437 @@
...
+        }
+    }
+
+    pub fn as_index(&self) -> Option<usize> {
+        match &self {
+            CirOperand::Index(index) => Some(*index),
+            CirOperand::Predicate(_p) => None, // never be here, otherwise something wrong
+        }
+    }
+}
```
Oyami-Srk, line 71: I would like to suggest using `panic!` instead of `Option` and then `unwrap` on the use case if `None` is impossible case for that. Like:
```Rust
pub fn as_index(&self) -> usize {
    match &self {
        CirOperand::Index(index) => *index,
        CirOperand::Predicate(_) => panic!("Call as_index on predicate Operand, CirProgram is wrong.")
    }
}
```

This will eliminate verbose `unwrap` usage and make a better panic message.

src/cir.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        if index >= lir.instructions.len() {
+            // end of LirProgram
+            break;
+        }
+    }
+    debug_assert_eq!(operator_stack.len(), 0);
+}
+
+#[inline]
```
ProBrian, line 261: Do we have tests for the explicitly inline version and non-explicitly inline version? That is, after we mark this funciton `inline`, what's the performance change(AFAIK, rust compiler may do inline itself without explicitly marking `inline` if it thinks that's an optimization), and is there binary inflation(since there are recursion in this function, I'm not sure how `inline` will expand the recursion logic)?

src/cir.rs
```diff
@@ -0,0 +1,437 @@
...
+                        index += 2;
+
+                        reduce_translation_stack(
+                            &mut cir.instructions,
+                            &mut operator_stack,
+                            &mut operand_stack,
+                        );
+                    }
+                }
+            },
```
Oyami-Srk, line 240: I would like to suggest to move out operator check on `next_ins` to reduce duplicated code.
And may using `BinaryIns` instead of `AndIns`/`OrIns` for better readibility.

```suggestion
            LirInstruction::LogicalOperator(op) => {
                let next_ins = &lir.instructions[index + 1];
                if is_operator(next_ins) {
                    // next is operator
                    operator_stack.push(*op); // Need LirLogicalOperators to implement Copy
                    index += 1;
                } else {
                    let (cir_instructions, operator_stack, operand_stack) = match op {
                        LirLogicalOperators::And | LirLogicalOperators::Or => {
                            let next_next_ins = &lir.instructions[index + 2];
                            cir.instructions.push(match op {
                                LirLogicalOperators::And => CirInstruction::AndIns(AndIns {
                                    left: CirOperand::Predicate(
                                        next_ins.as_predicate().unwrap().clone(),
                                    ),
                                    right: CirOperand::Predicate(
                                        next_next_ins.as_predicate().unwrap().clone(),
                                    ),
                                }),
                                LirLogicalOperators::Or => CirInstruction::OrIns(OrIns {
                                    left: CirOperand::Predicate(
                                        next_ins.as_predicate().unwrap().clone(),
                                    ),
                                    right: CirOperand::Predicate(
                                        next_next_ins.as_predicate().unwrap().clone(),
                                    ),
                                }),
                                _ => unreachable!(),
                            });
                            operand_stack.push(CirOperand::Index(cir.instructions.len() - 1));
                            index += 3;

                            (
                                &mut cir.instructions,
                                &mut operator_stack,
                                &mut operand_stack,
                            )
                        }
                        LirLogicalOperators::Not => {
                            next_ins.as_predicate().unwrap(); //right
                            let not_ins = NotIns {
                                right: CirOperand::Predicate(
                                    next_ins.as_predicate().unwrap().clone(),
                                ),
                            };
                            cir.instructions.push(CirInstruction::NotIns(not_ins));
                            operand_stack.push(CirOperand::Index(cir.instructions.len() - 1));
                            index += 2;

                            (
                                &mut cir.instructions,
                                &mut operator_stack,
                                &mut operand_stack,
                            )
                        }
                    };
                    reduce_translation_stack(cir_instructions, operator_stack, operand_stack);
                }
            },
```

src/cir.rs
xianghai2: fixed, thanks.

src/cir.rs
xianghai2: fixed, same thing. thanks.

src/cir.rs
xianghai2: the depth should be same.

src/interpreter.rs
xianghai2: fixed.

src/cir.rs
xianghai2: fixed.

src/cir.rs
xianghai2: fixed, good point. thanks.

src/cir.rs
xianghai2: fixed, typo. thanks.

src/cir.rs
xianghai2: fixed, good point, much better. thanks.

src/lir.rs
```diff
@@ -0,0 +1,465 @@
...
+
+            println!(
+                "The size of lir program: {} bytes",
+                mem::size_of::<LirProgram>()
+                    + mem::size_of::<LirInstruction>() * lir.instructions.len()
+            );
+            println!("The number of lir instructions: {}", lir.instructions.len());
+            println!("The lir instructions:");
+            println!("{:?}", lir.instructions);
+        }
```
Oyami-Srk, line 141: Should we move this into something like bench/UTs or else? Printing things to stdout seems like inappropriate for a library even if for he debug build. Or at least use `eprintln!` to print them to stderr.

src/cir.rs
```diff
@@ -0,0 +1,426 @@
...
+            use std::mem;
+            println!(
+                "The size of cir program: {} bytes",
+                mem::size_of::<CirProgram>()
+                    + mem::size_of::<CirInstruction>() * cir.instructions.len()
+            );
+            println!("The number of cir instructions: {}", cir.instructions.len());
+            println!("The cir instructions:");
+            println!("{:?}", cir.instructions);
+        }
```
Oyami-Srk, line 100: Same as `Translate::Expression`, I prefer not priniting to stdout directly.

src/lir.rs
```diff
@@ -0,0 +1,465 @@
...
+#[inline]
+pub(crate) fn is_operator(instruction: &LirInstruction) -> bool {
+    match instruction {
+        LirInstruction::LogicalOperator(_op) => true,
+        LirInstruction::Predicate(_p) => false,
+    }
+}
+
+#[derive(Debug)]
+pub enum LirLogicalOperators {
```
Oyami-Srk, line 48: Not using plural form is more appropriate.
```suggestion
pub enum LirLogicalOperator {
```

Oyami-Srk:
There is a way to get rid of `clone()` of `Predicate` inside Lir converting from Expression. Perhaps we can get rid of it too on Cir? `Predicate` is relatively large, not cloning it may result in better performance. 

[0001-try-to-get-rid-of-clone-of-expr-to-lir.patch](https://github.com/user-attachments/files/17001471/0001-try-to-get-rid-of-clone-of-expr-to-lir.patch)

src/cir.rs
```diff
@@ -0,0 +1,447 @@
...
+    ctx: &mut Context,
+    m: &mut Match,
+) -> bool {
+    match &cir_instructions[index] {
+        CirInstruction::AndIns(and) => {
+            let left_val = if is_index(&and.left) {
+                execute_helper(cir_instructions, and.left.as_index(), ctx, m)
+            } else {
+                and.left.as_predicate().execute(ctx, m)
+            };
```
Oyami-Srk, line 288: Why not just using `match`?

src/cir.rs
Oyami-Srk: Here is a better version if the latest commit does not include this. And this avoids panic when lir is empty.

```Rust
fn cir_translate_helper(lir: &LirProgram, cir: &mut CirProgram) {
    let mut operand_stack: Vec<CirOperand> = Vec::new();
    let mut operator_stack: Vec<LirLogicalOperators> = Vec::new();
    let mut index = 0;
    loop {
        if index >= lir.instructions.len() {
            // end of LirProgram
            break;
        }

        match &lir.instructions[index] {
            LirInstruction::LogicalOperator(op) => {
                let next_ins = &lir.instructions[index + 1];
                if is_operator(next_ins) {
                    // next is operator
                    operator_stack.push(*op); // Need LirLogicalOperators to implement Copy
                    index += 1;
                    continue;
                } else {
                    match op {
                        // Binary experssion
                        LirLogicalOperators::And | LirLogicalOperators::Or => {
                            let next_next_ins = &lir.instructions[index + 2];
                            cir.instructions.push(match op {
                                LirLogicalOperators::And => CirInstruction::AndIns(AndIns {
                                    left: CirOperand::Predicate(next_ins.as_predicate().clone()),
                                    right: CirOperand::Predicate(
                                        next_next_ins.as_predicate().clone(),
                                    ),
                                }),
                                LirLogicalOperators::Or => CirInstruction::OrIns(OrIns {
                                    left: CirOperand::Predicate(next_ins.as_predicate().clone()),
                                    right: CirOperand::Predicate(
                                        next_next_ins.as_predicate().clone(),
                                    ),
                                }),
                                _ => unreachable!(),
                            });
                            operand_stack.push(CirOperand::Index(cir.instructions.len() - 1));
                            index += 3;
                        }
                        // Unary expression
                        LirLogicalOperators::Not => {
                            next_ins.as_predicate(); //right
                            let not_ins = NotIns {
                                right: CirOperand::Predicate(next_ins.as_predicate().clone()),
                            };
                            cir.instructions.push(CirInstruction::NotIns(not_ins));
                            operand_stack.push(CirOperand::Index(cir.instructions.len() - 1));
                            index += 2;
                        }
                    };
                }
            }
            LirInstruction::Predicate(p) => {
                operand_stack.push(CirOperand::Predicate(p.clone()));
                index += 1;
            }
        }

        // End of a logical block
        reduce_translation_stack(
            &mut cir.instructions,
            &mut operator_stack,
            &mut operand_stack,
        );
    }
    debug_assert_eq!(operator_stack.len(), 0);
}
```

src/cir.rs
```diff
@@ -0,0 +1,447 @@
...
+        CirOperand::Index(_index) => true,
+        CirOperand::Predicate(_p) => false,
+    }
+}
+
+impl Translate for LirProgram {
+    type Output = CirProgram;
+    fn translate(&self) -> Self::Output {
+        let mut cir = CirProgram::new();
+        cir_translate_helper(self, &mut cir);
```
Oyami-Srk, line 92: Should we check if the `LirProgram` is empty?

xianghai2: fixed, check it in   cir_translate_helper().

src/cir.rs
xianghai2: Fixed.
The original intention was to reduce duplicated code and expect compiler could embed the inline hinted is_index here. However, it looks like the compiler seems not do inline is_index here, at least in debug mode for Linux/x86 platform on my dev pc.

src/lir.rs
xianghai2: fixed.

src/cir.rs
xianghai2: fixed.

src/lir.rs
xianghai2: fixed.

src/cir.rs
xianghai2: fixed.

src/lir.rs
```diff
@@ -0,0 +1,452 @@
...
+    fn translate(&self) -> Self::Output {
+        let mut lir = LirProgram::new();
+        lir_translate_helper(self, &mut lir);
+        lir.instructions.shrink_to_fit(); // shrink the memory
+        #[cfg(debug_assertions)]
+        {
+            let ast_counter = self.count_size();
+            ast_counter.ins_bytes;
+            ast_counter.ins_number;
+        }
```
Oyami-Srk, line 128: Please comment out the code without any effect. Clippy is unhappy with that. Otherwise, use `let _ = ...` to bypass the check (But it makes no sense to leave unused statements here.)

src/lir.rs
```diff
@@ -0,0 +1,452 @@
...
+    use crate::interpreter::Execute;
+    use crate::router::Router;
+    use crate::schema::Schema;
+    use crate::semantics::FieldCounter;
+    use std::collections::HashMap;
+    use uuid::Uuid;
+
+    #[derive(Debug, Clone, Copy)]
+    pub enum OperandItem<'a> {
+        Val(bool),
```
Oyami-Srk, line 173: Never constructed varient? Should us mark it as `dead_code`?

src/lir.rs
```diff
@@ -0,0 +1,452 @@
...
+        match item {
+            OperandItem::Val(b) => b,
+            OperandItem::Predicate(p) => p.execute(ctx, m),
+        }
+    }
+
+    #[inline]
+    fn check_short_circuit(operator_stack: &[LirLogicalOperator], operand_stack: &[bool]) -> bool {
+        // if it could be short-circuited, return true
+        if (operand_stack.len() > 0) && (operator_stack.len() > 0) {
```
Oyami-Srk, line 188: I'd like to suggest using `is_empty` to check and avoid direct length comparsion to make code more idiomatic and better for potential replacement of container type. Same to the rest of the code with `.len() > 0` or `.len() >= 1`

src/lir.rs
```diff
@@ -0,0 +1,452 @@
...
+                    let operand = operand_stack.last().unwrap();
+                    if *operand {
+                        return true;
+                    } else {
+                        return false;
+                    }
+                }
+                LirLogicalOperator::Not => {
+                    return false;
+                }
```
Oyami-Srk, line 208: Why not:

```suggestion
                LirLogicalOperator::And => {
                    !*operand_stack.last().unwrap()
                }
                LirLogicalOperator::Or => {
                    *operand_stack.last().unwrap()
                }
                LirLogicalOperator::Not => {
                    false
                }
```

src/lir.rs
```diff
@@ -121,6 +121,7 @@ impl Translate for Expression {
         lir_translate_helper(self, &mut lir);
         lir.instructions.shrink_to_fit(); // shrink the memory
         #[cfg(debug_assertions)]
+        #[allow(clippy::all)]
```
Oyami-Srk, line 124: I think we should take care of the usage of `allow` and any other explicit or implicit check bypassing.
I'd suggest any use of `#[allow]` should come with brief comments to explain why this `allow` is needed.

Oyami-Srk: @ADD-SP @dndx I suggest appending this to the Houdini coding standard to keep a better code quality.

src/lir.rs
xianghai2: fixed.

src/lir.rs
xianghai2: fixed.

src/lir.rs
xianghai2: fixed.  cleaned up legacy code as it is not needed now.

src/lir.rs
xianghai2: fixed

src/cir.rs
```diff
@@ -0,0 +1,451 @@
...
+    }
+}
+
+#[inline]
+fn is_index(cir_operand: &CirOperand) -> bool {
+    match cir_operand {
+        CirOperand::Index(_index) => true,
+        CirOperand::Predicate(_p) => false,
+    }
+}
```
StarlightIbuki, line 87: I'm surprised that Rust does not have a mechanism to assume or assert an enum variant.

src/cir.rs
```diff
@@ -0,0 +1,451 @@
...
+    Index(usize),
+    Predicate(Predicate),
+}
+
+impl CirOperand {
+    #[inline]
+    pub fn as_predicate(&self) -> &Predicate {
+        match &self {
+            CirOperand::Index(_index) => {
+                panic!("Call as_predicate on index Operand, CirProgram is wrong.")
```
StarlightIbuki, line 64:
```suggestion
                panic!("unexpected call to as_predicate with index Operand.")
```

(The error message should start with lower case.)

And the same for the following.

src/cir.rs
```diff
@@ -0,0 +1,451 @@
...
+
+#[derive(Debug)]
+pub struct NotIns {
+    right: CirOperand,
+}
+
+#[derive(Debug, Clone)]
+pub enum CirOperand {
+    Index(usize),
+    Predicate(Predicate),
```
StarlightIbuki, line 56: Shall we move the `predicate` to the same level as logic operators for more balanced sizes of nodes? This way it also saves memory usage.

src/cir.rs
```diff
@@ -0,0 +1,451 @@
...
+
+#[inline]
+fn is_index(cir_operand: &CirOperand) -> bool {
+    match cir_operand {
+        CirOperand::Index(_index) => true,
+        CirOperand::Predicate(_p) => false,
+    }
+}
+
+impl Translate for LirProgram {
```
StarlightIbuki, line 89: Can we translate directly from AST?

src/cir.rs
xianghai2: yes, it is defined inside enum "CirInstruction".

src/cir.rs
xianghai2: fixed, thanks

src/lir.rs
```diff
@@ -53,22 +53,22 @@ pub enum LirLogicalOperator {
...
 
-impl Translate for Expression {
-    type Output = LirProgram;
-    fn translate(&self) -> Self::Output {
+impl<'a> Translate<'a> for Expression {
+    type Output = LirProgram<'a>;
+    fn translate(&'a self) -> Self::Output {
         let mut lir = LirProgram::new();
         lir_translate_helper(self, &mut lir);
         lir.instructions.shrink_to_fit(); // shrink the memory
```
ProBrian, line 66: Just wondering if we need to shrink here, as I think `LirProgram` is just a intermedia state and will be abandoned as soon as `CirProgram` created, and shrink also consumes time here.

xianghai2: nice to have. and actually the memory of LirProgram is not returned to OS immediately.

src/cir.rs
xianghai2: Fixed, thanks.
Good point, now translate from AST drectly.

src/lir.rs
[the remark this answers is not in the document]
xianghai2: fixed, thanks.

src/cir.rs
Oyami-Srk: Because that assertion isn't necessary to be a part of the language itself. `match` and `unreachable` are enough to do this right. Perhaps some third-party crates may provide macros like `#[derive(unwrap_into_variant)]` to simplify this. But I don't think it's very needed for an external crate to do this simple job.

src/lir.rs
chobits: resolved I think

src/cir.rs
xianghai2: right, that could save a few bytes and it depends on the matcher pattern distribution. And, this will increase the memory access which degrades runtime performance.

src/cir.rs
```diff
@@ -0,0 +1,332 @@
...
+
+                let right = match r {
+                    Expression::Logical(_logic_exp) => {
+                        CirOperand::Index(cir_translate_helper(r, cir) - 1)
+                    }
+                    Expression::Predicate(p) => CirOperand::Predicate(p.clone()),
+                };
+                let and_ins = AndIns { left, right };
+                cir.instructions.push(CirInstruction::AndIns(and_ins));
+                cir.instructions.len()
```
ADD-SP, line 122: This can be moved out from the `match` statement.

src/cir.rs
```diff
@@ -0,0 +1,332 @@
...
+
+                let right = match r {
+                    Expression::Logical(_logic_exp) => {
+                        CirOperand::Index(cir_translate_helper(r, cir) - 1)
+                    }
+                    Expression::Predicate(p) => CirOperand::Predicate(p.clone()),
+                };
+                let or_ins = OrIns { left, right };
+                cir.instructions.push(CirInstruction::OrIns(or_ins));
+                cir.instructions.len()
```
ADD-SP, line 140: This can be moved out from the `match` statement.

src/cir.rs
```diff
@@ -0,0 +1,332 @@
...
+            LogicalExpression::Not(r) => {
+                let right: CirOperand = match r {
+                    Expression::Logical(_logic_exp) => {
+                        CirOperand::Index(cir_translate_helper(r, cir) - 1)
+                    }
+                    Expression::Predicate(p) => CirOperand::Predicate(p.clone()),
+                };
+                let not_ins = NotIns { right };
+                cir.instructions.push(CirInstruction::NotIns(not_ins));
+                cir.instructions.len()
```
ADD-SP, line 151: This can be moved out from the `match` statement.

src/cir.rs
```diff
@@ -0,0 +1,332 @@
...
+            }
+        },
+        Expression::Predicate(p) => {
+            cir.instructions.push(CirInstruction::Predicate(p.clone()));
+            cir.instructions.len()
```
ADD-SP, line 156: This can be moved out from the `match` statement.

src/ast.rs
```diff
@@ -136,9 +136,76 @@ pub struct Predicate {
...
+            match self {
+                Expression::Logical(l) => match l.as_ref() {
+                    LogicalExpression::And(l, r) => l.execute(ctx, m) && r.execute(ctx, m),
+                    LogicalExpression::Or(l, r) => l.execute(ctx, m) || r.execute(ctx, m),
+                    LogicalExpression::Not(r) => !r.execute(ctx, m),
+                },
+                Expression::Predicate(p) => p.execute(ctx, m),
+            }
+        }
+    }
```
ADD-SP, line 207: This has been moved to the tests module, do we have any tests are using this?

benches/build.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    let mut context = Context::new(&schema);
+    context.add_value("a", Value::Int(N as i64));
+
+    c.bench_function("Build Router", |b| {
+        b.iter_with_large_drop(|| {
+            let mut router = Router::new(&schema);
+            for i in 0..N {
+                let expr = format!("((a > 0 || a < {}) && a != 0) && a == 1", N + 1);
+                router.add_matcher(N - i, Uuid::new_v4(), &expr).unwrap();
```
ADD-SP, line 27: This line doesn't work.

src/cir.rs
xianghai2: fixed.

src/cir.rs
xianghai2: fixed.

src/cir.rs
xianghai2: fixed.

src/ast.rs
xianghai2: fixed.

benches/build.rs
xianghai2: fixed, thanks.

src/ast.rs
```diff
@@ -6,6 +6,34 @@ use std::net::IpAddr;
...
+        Self { stack: Vec::new() }
+    }
+}
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub enum RouteTerm {
+    LogicalOperator(RouteLogicalOperators),
+    Predicate(Predicate),
+}
```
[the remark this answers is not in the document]
chobits: resolved, RouteTerm has been modified. so i think we could resovle this conversation

src/cir.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+    cir.instructions.len()
+}
+
+fn execute_helper(
+    cir_instructions: &[CirInstruction],
+    index: usize,
+    ctx: &mut Context,
+    m: &mut Match,
+) -> bool {
```
StarlightIbuki, line 163: Consider using a slice instead of "reference and index"

src/cir.rs
```diff
@@ -0,0 +1,439 @@
...
+                }
+            }
+        } else {
+            // operator stack drained
+            break;
+        }
+    }
+}
+
+fn cir_translate_helper(lir: &LirProgram, cir: &mut CirProgram) {
```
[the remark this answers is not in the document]
chobits: I found @xianghai2 has written one test to run `ast.translate()` function, which will trigger the `cir_translate_helper `function

src/cir.rs
StarlightIbuki: You can make a recursive call like this: `execute_helper(&cir_instructions[index..], ctx, m)`

src/cir.rs
```diff
@@ -0,0 +1,365 @@
...
+    cir.instructions.len()
+}
+
+fn execute_helper(
+    cir_instructions: &[CirInstruction],
+    index: usize,
+    ctx: &mut Context,
+    m: &mut Match,
+) -> bool {
+    match &cir_instructions[index] {
```
StarlightIbuki, line 164: I recommend reordering the instructions. Reading the array in inverse order hurts the CPU's caching performance.

src/cir.rs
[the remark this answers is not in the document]
chobits: for new function 

```
fn cir_translate_helper(exp: &Expression, cir: &mut CirProgram) -> usize {
```

Ok, i have reviewed the test cases, they only check the assertion result, but not check if the `exp` has been converted to expected `cir` structure. I think we could resovlve this conversation, then left a TODO(kag) to write a helper function to reverse cir back to exp string to do the verification in UTs.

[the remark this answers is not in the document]
chobits: open a ticket for me to practice rust: https://konghq.atlassian.net/browse/KAG-5462

src/cir.rs
StarlightIbuki: Never mind. I thought you intend to pass a slice.

StarlightIbuki (approved):
Though there are potential improvement can be done, the current implementation is okay to me.

ProBrian (approved):

Oyami-Srk (approved):
LGTM, clippy will be happy after merging/rebase.

ADD-SP (approved):
LGTM

nowNick (commented):
Hey! @xianghai2 - could you attach some benchmarking reports to this PR? I guess just run the benches on main branch and here to see the impact of this optimization.

src/ast.rs
```diff
@@ -22,14 +22,14 @@ pub enum LogicalExpression {
 }
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
dndx, line 25: This is a very small data structure, can it be `Copy` as well?

src/cir.rs
```diff
@@ -0,0 +1,365 @@
...
+        match &self {
+            CirOperand::Index(_index) => {
+                panic!("unexpected call to as_predicate with index operand.")
+            }
+            CirOperand::Predicate(p) => p,
+        }
+    }
+
+    #[inline]
+    pub fn as_index(&self) -> usize {
```
dndx, line 70: Unused function?

src/cir.rs
```diff
@@ -0,0 +1,365 @@
...
+                };
+                let not_ins = NotIns { right };
+                cir.instructions.push(CirInstruction::NotIns(not_ins));
+            }
+        },
+        Expression::Predicate(p) => {
+            cir.instructions.push(CirInstruction::Predicate(p.clone()));
+        }
+    }
+    cir.instructions.len()
```
dndx, line 155:
```suggestion
    cir.instructions.len() - 1
```

Wouldn't this be more natural? Returns the index of translated IR.

We should document this function's argument and return value as well.

dndx (changes requested):

src/cir.rs
xianghai2: fixed.

src/cir.rs
xianghai2: fixed, remove it.

src/ast.rs
xianghai2: fixed.

Oyami-Srk (changes requested):

src/cir.rs
```diff
@@ -0,0 +1,361 @@
...
+        execute_helper(&self.instructions, self.instructions.len() - 1, ctx, m)
+    }
+}
+
+impl FieldCounter for CirProgram {
+    fn add_to_counter(&self, map: &mut HashMap<String, usize>) {
+        for instruction in &self.instructions {
+            match &instruction {
+                CirInstruction::AndIns(and) => {
+                    if !is_index(&and.left) {
```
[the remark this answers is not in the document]
xianghai2: fixed, implemented FieldCounter for CirOperand.

src/cir.rs
```diff
@@ -0,0 +1,361 @@
...
+                        *map.entry(and.left.as_predicate().lhs.var_name.clone())
+                            .or_default() += 1;
+                    }
```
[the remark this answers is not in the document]
xianghai2: fixed.

Oyami-Srk (approved):

src/cir.rs
```diff
@@ -0,0 +1,311 @@
...
+            }
+        }
+    }
+}
+
+impl FieldCounter for CirProgram {
+    fn add_to_counter(&self, map: &mut HashMap<String, usize>) {
+        for instruction in &self.instructions {
+            instruction.add_to_counter(map);
+        }
```
dndx, line 246: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

This would be cleaner.

dndx (changes requested):

src/cir.rs
xianghai2: fixed

xianghai2:
> Hey! @xianghai2 - could you attach some benchmarking reports to this PR? I guess just run the benches on main branch and this branch to see the impact of this optimization.

@nowNick you may refer to https://konghq.atlassian.net/browse/KAG-5334 for benchmark details. thanks.

dndx (approved):

PR #247, opened 2024-09-04
use flat array for field indexing instead of hashmap

ProBrian:
To resolve [KAG-5155](https://konghq.atlassian.net/browse/KAG-5155)

### DIFF
--- a/lib/resty/router/context.lua
+++ b/lib/resty/router/context.lua
@@ -26,20 +26,25 @@ local clib = cdefs.clib
 local context_free = cdefs.context_free
 
 
-function _M.new(schema)
-    local context = clib.context_new(schema.schema)
+function _M.new(schema, size)
+    local context = clib.context_new(size)
     local c = setmetatable({
         context = ffi_gc(context, context_free),
         schema = schema,
+        max_fields = size,
     }, _MT)
 
     return c
 end
 
 
-function _M:add_value(field, value)
+function _M:add_value(index, field, value)
     if not value then
         return true
+    end
+
+    if index > self.max_fields then
+        return false, "context value index out of bound"
     end
 
     local typ, err = self.schema:get_field_type(field)
@@ -65,7 +70,7 @@ function _M:add_value(field, value)
     local errbuf_len = get_size_ptr()
     errbuf_len[0] = ERR_BUF_MAX_LEN
 
-    if clib.context_add_value(self.context, field, CACHED_VALUE, errbuf, errbuf_len) == false then
+    if clib.context_add_value(self.context, index-1, CACHED_VALUE, errbuf, errbuf_len) == false then
         return nil, ffi_string(errbuf, errbuf_len[0])
     end
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,5 +1,4 @@
 use crate::ast::Value;
-use crate::schema::Schema;
 use fnv::FnvHashMap;
 use uuid::Uuid;
 
@@ -25,38 +24,73 @@ impl Default for Match {
     }
 }
 
-pub struct Context<'a> {
-    schema: &'a Schema,
-    values: FnvHashMap<String, Vec<Value>>,
+pub struct Context {
+    values: Vec<Option<Vec<Value>>>,
     pub result: Option<Match>,
 }
 
-impl<'a> Context<'a> {
-    pub fn new(schema: &'a Schema) -> Self {
+impl Context {
+        pub fn new(fields_cnt: usize) -> Self {
         Context {
-            schema,
-            values: FnvHashMap::with_hasher(Default::default()),
+            values: vec![None; fields_cnt],
             result: None,
         }
     }
 
-    pub fn add_value(&mut self, field: &str, value: Value) {
-        if &value.my_type() != self.schema.type_of(field).unwrap() {
-            panic!("value provided does not match schema");
+    pub fn add_value(&mut self, index: usize, value: Value) {
+        if index >= self.values.len() {
+            panic!("value provided does not match schema: index {}, max fields count {}", index, self.values.len());
         }
 
-        self.values
-            .entry(field.to_string())
-            .or_default()
-            .push(value);
+        if let Some(v) = &mut self.values[index] {
+            v.push(value);
+        } else {
+            self.values[index] =  Some(vec![value]);
+        }
     }
 
-    pub fn value_of(&self, field: &str) -> Option<&[Value]> {
-        self.values.get(field).map(|v| v.as_slice())
+    pub fn value_of(&self, index: usize) -> Option<&[Value]> {
+        if !self.values.is_empty() && self.values[index].is_some() { Some(self.values[index].as_ref().unwrap().as_slice()) } else {None}
     }
 
     pub fn reset(&mut self) {
+        let len = self.values.len();
+        // reserve the capacity of values for reuse, avoid re-alloc
         self.values.clear();
+        self.values.resize_with(len, Default::default);
         self.result = None;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::context::Context;
+    use crate::ast::Value;
+    #[test]
+    fn test_context() {
+        let fields_cnt = 3;
+        let mut ctx = Context::new(fields_cnt);
+        assert!(ctx.values.len() == fields_cnt);
+        assert_eq!(ctx.values, vec![None; fields_cnt]);
+        // access value with out of bound index
+        assert_eq!(ctx.value_of(0), None);
+        
+        // add value in bound
+        ctx.add_value(1, Value::String("foo".to_string()));
+        assert_eq!(ctx.value_of(0), None);
+        assert_eq!(ctx.value_of(1).unwrap().len(), 1);
+        assert_eq!(ctx.value_of(1).unwrap(), vec![Value::String("foo".to_string())].as_slice());
+
+        // reset context keeps values capacity with all None 
+        ctx.reset();
+        assert!(ctx.values.len() == fields_cnt);
+        assert_eq!(ctx.values, vec![None; fields_cnt]);
+
+        // reuse this context
+        ctx.add_value(0, Value::String("bar".to_string()));
+        ctx.add_value(0, Value::String("foo".to_string()));
+        assert!(ctx.values.len() == fields_cnt);
+        assert_eq!(ctx.value_of(0).unwrap().len(), 2);
+        assert_eq!(ctx.value_of(0).unwrap(), vec![Value::String("bar".to_string()), Value::String("foo".to_string())].as_slice());
     }
 }
--- a/src/router.rs
+++ b/src/router.rs
@@ -13,7 +13,8 @@ struct MatcherKey(usize, Uuid);
 pub struct Router<'a> {
     schema: &'a Schema,
     matchers: BTreeMap<MatcherKey, Expression>,
-    pub fields: HashMap<String, usize>,
+    pub fields: Vec<(String, usize)>, // fileds array of tuple(name, count)
+    pub fields_map: HashMap<String, usize>, // field name -> index map
 }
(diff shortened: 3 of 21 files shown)


src/context.rs
```diff
@@ -25,38 +24,73 @@ impl Default for Match {
     }
 }
 
-pub struct Context<'a> {
-    schema: &'a Schema,
```
ProBrian, line 29: `Schema` is no more needed here, since `Schema` uses field name as hash key and `Context` no longer uses field name to add value, but only index in `Router`'s fields array

src/context.rs
```diff
@@ -25,38 +24,73 @@ impl Default for Match {
...
-    values: FnvHashMap<String, Vec<Value>>,
+pub struct Context {
+    values: Vec<Option<Vec<Value>>>,
```
ProBrian, line 28: uses `values` array to store all field values, `values` index match the index of `Router`'s `fields` array

src/router.rs
```diff
@@ -13,15 +13,17 @@ struct MatcherKey(usize, Uuid);
 pub struct Router<'a> {
     schema: &'a Schema,
     matchers: BTreeMap<MatcherKey, Expression>,
-    pub fields: HashMap<String, usize>,
+    pub fields: Vec<(String, usize)>, // fileds array of tuple(name, count)
+    pub fields_map: HashMap<String, usize>, // field name -> index map
```
ProBrian, line 17: `fields` is an flat array to store all field names and corresponding counts.
Here we uses an hashmap `fields_map` to store the field's index in `fields` array, `fields_map` is not used in runtime, but only for router matcher add and delete, when matcher changed, use this map to quickly get the index of fields.

lib/resty/router/context.lua
```diff
@@ -26,22 +26,27 @@ local clib = cdefs.clib
...
         schema = schema,
+        max_fields = size,
     }, _MT)
 
     return c
 end
 
 
-function _M:add_value(field, value)
+function _M:add_value(index, field, value)
```
ADD-SP, line 41: Does this mean the caller has to maintain a map from `field_name` to `index`?

Oyami-Srk: I am curious about this too. Could we use something like `HashMap<String, usize>` from `Route` to do the internal field name to index map instead of exposing the internal index to the outside world? But this sounds like making the whole improvement meaningless.

Or could we update `Schema` to not only save the type of a field but also save the calculated index from `Router`? I do not fully understand the ATC-Router, correct me if that's impossible.

Currently, I am aware that We got three primary different components of ATC-Router, which are `Route`, `Schema`, and `Context`. `Route` is created from `Schema` and performs execution on `Context`, in a relative loose manner, `Context` is related with `Schema` but not with `Route`. 

Is this necessary to not create `Context` from `Route` but from nothing? The former solution will allow us to integrate the index onto `Context` and will allow us to use an updated `Schema::fields`, like `Route:fields` but with an index inside the map. Since the HashMap lookup is not avoided inside `self.schema:get_field_type(field)`, I think it is better to get the index from it as well.

ProBrian: The caller uses `router:get_fields` to maintain an array of fields, here is the usage in kong:
```
function _M:fields_visitor(params, ctx, cb, cb_arg)
  for idx, field in ipairs(self.fields) do
    local value = self:get_value(field, params, ctx)
    -- add_value invoked in the callback
    local res, err = cb(field, value, cb_arg, idx)
    if not res then
      return nil, err
    end
  end -- for fields

  return true
end
```
This is a loose couple between `Context::values` and `Router::fields`, only using index to match.

ProBrian: > Is this necessary to not create `Context` from `Route` but from nothing? The former solution will allow us to integrate the index onto `Context` and will allow us to use an updated `Schema::fields`, like `Route::fields` but with an index inside the map. Since the HashMap lookup is not avoided inside `self.schema:get_field_type(field)`, I think it is better to get the index from it as well.

Yes the previous `Context` implementation keeps `Schema`, which is used for type validation. But noticed that `Schema::fields` is different from `Router::fields`, `Router::fields` and exactly used field in configured router rules, like `['http.path', 'http.header.foo', 'http.header.bar']`, but `Schema::fields` maintains fixed schema of field, which like: `{'http.path': 'string', 'http.headers.*': 'string', 'net.src.port': 'ipaddr'}`, so there are not one-one mapping between router fields and schemas. 
in previous version, we keep schema in `Context` to do the field type validation, and for now, since we don't pass field name into `Context` anymore, the schema is useless now. However, We still check the field type with schema in the Lua side, before invoking `add_value`.

ADD-SP: > This is a loose couple between Context::values and Router::fields, only using index to match.

As the end developer, I don't want to maintain this map by myself, this should be done inside the crate.

ProBrian: > > This is a loose couple between Context::values and Router::fields, only using index to match.
> 
> As the end developer, I don't want to maintain this map by myself, this should be done inside the crate.

Sure, if we consider the ease of use, and robustness of this module, it's better to maintain the router fields inside the `Context`, assume that we have an `fields`(map of name: index) inside the context, and we provide the API to end user like `add_value(field, value)`, then the implementation logic maybe like:
```
function _M:add_value(field_name, value)
    local index = self.fields[field_name] -- an hash find occurs here.
    ...
    clib.context_add_value(index, ...)
```
And what this KAG tries to do is to avoid unnecessary hash find to improve performance, which is inspired by the implementation of nginx's variable indexed fields mechanism. 

if we map to nginx, the `Context` is basically like `r->variables`, which stores a flat array of variable values(without storing the variable name to value, or name to index mapping inside). what `r->variables` only knows, is it can use index to get/set variable from an global structure `cmcf`(which likes the role `Router::fields`). 

`Context` and `r->variables` are more likely to represent some runtime state which changes frequently, we want to transfer and access them as fast as possible(and as smaller as we can). That makes them looks not so "easy to use", but I think that's a trade-off which worth it.

Oyami-Srk: What if we can store a reference of used `Schema::fields` along with those actual index values used in `Route` to some field inside `Context` and invoke HashMap search here, to get the index and type both?

`fields_visitor` invokes `add_value` eventually which means the HashMap lookup still happens. Is there any chance to avoid exposing the index by linking the `Context` and `Route` together?

ProBrian: > What if we can store a reference of used `Schema::fields` along with those actual index values used in `Route` to some field inside `Context` and invoke HashMap search here, to get the index and type both?
> 
> `fields_visitor` invokes `add_value` eventually which means the HashMap lookup still happens. Is there any chance to avoid exposing the index by linking the `Context` and `Route` together?

The previous implementation has several hashmap finds:

- `Context::add_value()` : 
    - find `schema` to validate field type by field name(which I think is unnecessary because we already check in lua side)
    - insert value into `self.values` hashmap by field name
- `Expression::execute()`:
    - find `context.values` by `lhs`'s field name(which is the critical path that will be executed a lot)
    
I think our initial is to avoid such "heavy" computation as much as possible, that's why I remove hashmap in `Context` and replace `self.values` with a vector, assume that users get a fields list already, then what we can achieve is:

1. field name is no long needed in `Context`'s `self.values`, this reduces `Context`'s size.
2. insert value into `self.values` no longer need hash find, just vector indexing.
3. find value in `Expression` executing no long need hash find, just vector indexing.

That would bring considerable performance improvement in runtime.

However, since we remove field name in `Context`, we may concern that is if we are using those APIs concurrently(e.g. router fields changed after context values constructed but matching not yet executed), we may have wrong field matching(router fields index changed but context is not aware of). 
If we MUST guarantee this condition happens, maybe we should use a version in context to match router version(version in `Context`, `Router`, maybe also in `Expression`, that makes the version invasive to every data structure). Or we can change the `Context.values` to something like `Vec<struct {field_name, field_index, counter}>`, and making sure field name matches the `lhs.name` in every matching execution(however, string comparison also takes time).

src/context.rs
```diff
@@ -26,37 +25,165 @@ impl Default for Match {
 }
 
 pub struct Context<'a> {
-    schema: &'a Schema,
-    values: FnvHashMap<String, Vec<Value>>,
+    router: &'a Router<'a>,
```
ProBrian, line 28: bind `Context` with `Router` instead, since `Router` can provide more precise fields validation.

lib/resty/router/context.lua
Oyami-Srk: > find schema to validate field type by field name(which I think is unnecessary because we already check in lua side)
I believe we checked the type twice for different intents. The Lua side type check guarantees the value passed into the Rust side is always valid, and the Rust side checks for that validity. 
Should we just remove the type check on the Lua side and make the Rust side return an error if the type is invalid to reduce the redundant call of the type check? I am not sure if there are some special concerns.

I do see newly added code has internal index storage with `Router::fields`. But since the Schema is *possibly* always outlived with `Router`, could we use string references from `Schema` to build the fields by linking the `Schema` with the `Router`? Since we already did link `Context` with `Schema` on the Lua side. I am not very sure about that.

The current implementation already looks good to me.

dndx:
@ProBrian This PR need to be rebased

ProBrian:
Local benchmark result attached (MacOS M3 chip with 36GB mem), comparing with PR branch and main branch
```
 Running benches/build.rs (target/release/deps/build-3f0a7ade35c0c7a4)
New: Build Router            time:   [3.0939 ms 3.1021 ms 3.1114 ms]
Old: Build Router            time:   [3.1140 ms 3.1375 ms 3.1765 ms]
                        change: [+1.2814% +2.1473% +3.3212%] (p = 0.00 < 0.05)

 
     Running benches/match_mix.rs (target/release/deps/match_mix-8ebfb4b8ad1a6fd2)
New: Match                   time:   [36.678 ns 36.807 ns 36.991 ns]
Old: Match                   time:   [56.656 ns 56.897 ns 57.178 ns]
                        change: [+10.838% +11.498% +12.176%] (p = 0.00 < 0.05)

 
     Running benches/misc_match.rs (target/release/deps/misc_match-ef04ce8d11de66a9)
  New: route match all         time:   [103.81 ns 104.00 ns 104.17 ns]
  Old:  route match all         time:   [89.362 ns 89.910 ns 90.633 ns]

 
New: route mismatch all      time:   [123.28 ns 123.52 ns 123.80 ns]
Old: route mismatch all      time:   [225.65 ns 226.23 ns 226.94 ns]

 
New: route matchers batch create and delete
                        time:   [20.357 ms 20.448 ms 20.555 ms]
Old: route matchers batch create and delete
                        time:   [19.954 ms 20.100 ms 20.272 ms]

 
     Running benches/not_match_mix.rs (target/release/deps/not_match_mix-8894a45acea74fc2)
New: Doesn't Match           time:   [13.551 ms 13.624 ms 13.715 ms]
Old: Doesn't Match           time:   [21.577 ms 22.097 ms 22.635 ms]

 
     Running benches/string.rs (target/release/deps/string-4f7ea951a6e3683d)
New: Doesn't Match           time:   [10.536 ms 10.571 ms 10.615 ms]
Old: Doesn't Match           time:   [14.018 ms 14.069 ms 14.125 ms]

 
     Running benches/test.rs (target/release/deps/test-1f097a0352e64a6c)
New: Doesn't Match           time:   [35.154 ms 35.342 ms 35.552 ms]
Old: Doesn't Match           time:   [35.460 ms 35.740 ms 36.037 ms]

```

ProBrian:
@dndx  @ADD-SP Are there any new comments on this PR?

StarlightIbuki:
Please resolve conflicts

ProBrian:
> Please resolve conflicts

Conflicts caused due to the revert introduced by https://github.com/Kong/atc-router/pull/284, we should figure out if that reverted change will be committed back after the investigation of performance.

PR #251, opened 2024-09-19
perf(*): cache regex predicates with rc using router attribute

nowNick:
KAG-3182


ADD-SP (commented):
I prefer this solution:

1. LRU cache makes the performance less stable.
2. The thread_local solution should save more memory for more than one router instance, this is not the typical usage of the router, and I personally don't like keeping a global state in a standalone library.

I think this approach is fine, I'd like to discuss one thing. The current implementation leaks the caching logical during the parsing, It works, but it is not as elegant as it could be for the type designing as AST should just describe the syntax. Do you have any idea to make it better?

Maybe we can just store the regex string in AST and build that later.

bungle:
Do we need to take into account things like this:
https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#lua_regex_cache_max_entries

Aka in Lua PCRE regex caches we have a upper bound.

ADD-SP:
> Do we need to take into account things like this: https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#lua_regex_cache_max_entries
> 
> Aka in Lua PCRE regex caches we have a upper bound.

Let's say the customer configured 10 regexes, and our upper bound is 5, which means that we have to build regex on matching; it slows down the router-matching.

I think the reason OpenResty sets an upper bound is that OpenResty doesn't know the lifetime of each regex, but it doesn't want to build a regex each time, so it has to set an upper bound. 

For atc-router, we know the lifetime of each regex, so the upper bound is not required.

It makes sense to build *N* regexes in memory if the customer configured *N* regexes, it is an acceptable cost and makes the matching speed more stable.

ADD-SP:
> Do we need to take into account things like this: https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#lua_regex_cache_max_entries
> 
> Aka in Lua PCRE regex caches we have a upper bound.

@bungle 

Let's say the customer configured 10 regexes, and our upper bound is 5, which means that we have to build regex on matching; it slows down the router-matching.

I think the reason OpenResty sets an upper bound is that OpenResty doesn't know the lifetime of each regex, but it doesn't want to build a regex each time, so it has to set an upper bound. 

For atc-router, we know the lifetime of each regex, so the upper bound is not required.

It makes sense to build *N* regexes in memory if the customer configured *N* regexes, it is an acceptable cost and makes the matching speed more stable.

src/parser.rs
```diff
@@ -195,16 +204,26 @@ fn parse_predicate(pair: Pair<Rule>) -> ParseResult<Predicate> {
...
+
+                        let rc = Rc::new(r);
 
-                Value::Regex(r)
+                        regex_cache.insert(s, rc.clone());
+                        rc
+                    }
+                };
+
+                Value::Regex(regex_rc)
```
ADD-SP, line 226: We could use https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry to simply the logic.

src/router.rs
```diff
@@ -32,7 +36,7 @@ impl<'a> Router<'a> {
             return Err("UUID already exists".to_string());
         }
 
-        let ast = parse(atc).map_err(|e| e.to_string())?;
+        let ast = parse(atc, &mut self.regex_cache).map_err(|e| e.to_string())?;
```
dndx, line 39: Nice work, however, it seems that we never check to see if `Rc` gets down to `1` and regex should be released. We probably need to check the `Rc` count in `remove`, and if it gets to `1` (which means only `HashMap` is holding it), remove the entry.

dndx (changes requested):

src/parser.rs
nowNick: Actually... I don't think we can :/ 
I've wanted to use `entry` but it consumes string. I'd have to clone it but then it defeats the purpose of this optimization.
There's this big thread to that tried to address this by COW and lookup by reference but I don't think it got anywhere: https://github.com/rust-lang/rfcs/pull/1769

src/router.rs
nowNick: Huh! I exploring if there's a way to register a hook to `Rc` that will get triggered on `drop` but it does not look like there's anything. But you're right - we don't need something as sophisticated and complicated as that. The action of removing a matcher is a great place (even better as it's explicit) to do such removal. Will do that 👍

src/parser.rs
ADD-SP: @nowNick Ah, you are right, the `entry()` consumes the key.

src/parser.rs
```diff
@@ -195,16 +204,26 @@ fn parse_predicate(pair: Pair<Rule>) -> ParseResult<Predicate> {
...
-                    ParseError::new_from_span(
-                        ErrorVariant::CustomError {
-                            message: e.to_string(),
-                        },
-                        rhs_pair.as_span(),
-                    )
-                })?;
+                let regex_rc = match regex_cache.get(&s) {
+                    Some(stored_regex_rc) => stored_regex_rc.clone(),
+                    _ => {
```
ADD-SP, line 209:
```suggestion
                    None => {
```

For better semantics.

src/router.rs
nowNick: Fixed ✅

src/parser.rs
nowNick: Fixed ✅

nowNick:
Hey @ADD-SP :) !
I was thinking if there's a way to refactor it but unfortunately this is the only solution that comes to my mind. We have to find the first place that can hold an object that will be shared across `add_matcher` calls and that's the Router. We cannot store it in `ast` since it creates a new instance on each `add_matcher` call. Actually when you think of it `regex_cache` shares some similar logic to `fields`. They both store general information about the matchers that are shared across `add_matcher` calls. The first one remembers regexes the other one predicates.

The only other approaches I saw were in the other PRs linked in the description at the top.

src/router.rs
```diff
@@ -14,6 +16,27 @@ pub struct Router<'a> {
...
+        Expression::Logical(l) => match l.as_ref() {
+            LogicalExpression::And(l, r) | LogicalExpression::Or(l, r) => {
+                release_cache(l, router);
+                release_cache(r, router);
+            }
+            LogicalExpression::Not(r) => release_cache(r, router),
+        },
+        Expression::Predicate(p) => {
+            if let Value::Regex(rc) = &p.rhs {
+                if Rc::strong_count(&rc) == 2 {
```
ADD-SP, line 33:
```suggestion
                if Rc::strong_count(rc) == 2 {
```

截屏2024-10-08 10 08 37

nowNick: Fixed! Thanks ✅

src/router.rs
dndx: Yes, we try to make things as predictable as possible in Rust, so checking it only at the place it should be checked is preferred.

dndx:
@nowNick Could you rebase this PR?

nowNick:
@ADD-SP I need a little bit more time to resolve it. The changes with CPU locality optimizations are a little bit tricky to incorporate in this PR.

src/router.rs
```diff
@@ -1,19 +1,38 @@
...
 
 pub struct Router<'a> {
     schema: &'a Schema,
     matchers: BTreeMap<MatcherKey, CirProgram>,
     pub fields: HashMap<String, usize>,
+    pub regex_cache: HashMap<String, Rc<Regex>>,
+}
+
+fn release_cache(cir: &CirProgram, router: &mut Router) {
+    cir.instructions.iter().for_each(|instruction| {
```
nowNick, line 26: Since instructions are now just in an array I can simply iterate through it and release regexes when I encounter them. Previously I had to traverse the tree of AST.

nowNick:
I've rebased the branch to use CIR program. From what I can see the memory improvents remained the same ~19times less memory for this benchmark but what is more we achieved further CPU optimizations. From 80ms to 45ms! (46% better)

dndx:
@xianghai2 @ADD-SP Could you help @nowNick review this once you have some chance? Not a release blocker.

.github/workflows/tests.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   tests:
     name: Tests
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
```
nowNick, line 17: I had to upgrade the runner due to 20.04 deprecation.
https://github.com/actions/runner-images/issues/11101

lucab (approved):
Changes LGTM.
(There is a bit of re-inventing of `Weak` in the removing logic, but I think it makes sense as they would be slightly annoying to use here anyway)

Just pointing out that this changes some public signatures, so it would need a v2 bump. I'm not sure how those are coordinated here.

src/ffi/context.rs
```diff
@@ -158,8 +158,8 @@ pub unsafe extern "C" fn context_reset(context: &mut Context) {
 /// Violating any of the following constraints will result in undefined behavior:
 ///
 /// - `context` must be a valid pointer returned by [`context_new`],
-///    must be passed to [`router_execute`] before calling this function,
-///    and must not be reset by [`context_reset`] before calling this function.
+///   must be passed to [`router_execute`] before calling this function,
```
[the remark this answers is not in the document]
nowNick: Fixed in another PR ✅

.github/workflows/tests.yml
nowNick: Fixed in another PR ✅

src/ast.rs
```diff
@@ -53,7 +53,7 @@ pub enum Value {
     IpAddr(IpAddr),
     Int(i64),
     #[cfg_attr(feature = "serde", serde(with = "serde_regex"))]
-    Regex(Regex),
+    Regex(Rc<Regex>),
```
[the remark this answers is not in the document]
ADD-SP: Will `Rc` makes the entire router `!Send`?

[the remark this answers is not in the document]
lucab: I think so.
And the new `regex_cache: HashMap<String, Rc<Regex>>` field will definitely have that effect too.
Does this need to be used in a multi-threaded context?
(I didn't see any static assertions about Send/Sync)

ProBrian:
LGTM

lucab (approved):

PR #262, opened 2024-09-24
Eliminate Clippy warnings

ADD-SP:
## REBASE and MERGE

This PR eliminates all clippy warnings in the source code.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -63,11 +63,42 @@ pub extern "C" fn schema_new() -> *mut Schema {
     Box::into_raw(Box::default())
 }
 
+/// Deallocate the schema object.
+///
+/// # Errors
+///
+/// This function never fails.
+///
+/// # Safety
+///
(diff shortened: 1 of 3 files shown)


src/ffi.rs
```diff
@@ -126,11 +248,73 @@ pub unsafe extern "C" fn router_remove_matcher(
...
+///
+/// # Arguments
+///
+/// - `router`: a pointer to the [`Router`] object returned by [`router_new`].
+/// - `context`: a pointer to the [`Context`] object.
+///
+/// # Returns
+///
+/// Returns `true` if the router was executed successfully,
+/// `false` for no match found.
```
Oyami-Srk, line 261: I think the "execution" is always successful, otherwise we panic. I think the boolean returning from this interface is something like an execution result.

ADD-SP: Fixed

Oyami-Srk (approved):
LGTM

dndx (approved):

PR #263, opened 2024-09-24
feat(ffi): add FFI interface to get operators along with fields while performing expressions validation

Oyami-Srk:
This interface allows the caller to validate an expression string with schema and get its operators/fields without needing a Router instance.

KAG-5013


src/ffi.rs
```diff
@@ -578,6 +578,189 @@ pub unsafe extern "C" fn context_get_result(
...
+
+        for k in expr_fields.into_keys() {
+            total_fields_length += k.as_bytes().len() + 1; // +1 for trailing \0
+            fields.push(k);
+        }
+
+        if !fields_buf.is_null() {
+            if *fields_len < total_fields_length {
+                let err_msg = format!(
+                    "Fields buffer too small, provided {} bytes but required at least {} bytes.",
```
dndx, line 703: Since the validation result is used by a library, for these kind of "internal errors" that the user should not see, we should return a more structured error instead of the string, that way the caller knows what happened and can reallocate if desired.

I suggest changing the return of this FFI function to a `int` and define a couple of constants instead.

For errors that the caller can't do much and has to propagate to the user, a string error is okay.

dndx (changes requested):

src/ffi.rs
Oyami-Srk: Fixed, thanks.

src/ffi.rs
```diff
@@ -578,6 +578,199 @@ pub unsafe extern "C" fn context_get_result(
...
+        let errlen = min(e.len(), *errbuf_len);
+        errbuf[..errlen].copy_from_slice(&e.as_bytes()[..errlen]);
+        *errbuf_len = errlen;
+        return ATC_ROUTER_EXPRESSION_VALIDATE_FAILED;
+    }
+    // Unwrap is safe since we've already checked for error
+    let ast = result.unwrap();
+
+    // Validate expression with schema
+    if let Err(e) = ast.validate(schema).map_err(|e| e.to_string()) {
```
ProBrian, line 683: Maybe we could wrap `parse`, `validate` and `add_to_counter` logic in a single function, so that it could be call for both `Router` and `expression_validate`. 
The reason I mention this is that there maybe data structure change for above functions(e.g. I may change `add_to_counter`'s parameter in my PR for indexed fields), wrapping the implementaion logic in a single function would bring the benefit that we don't need to change every caller's code in future.

ProBrian: I just noticed that here we use `ast.add_to_counter`, and when Hai's PR merged, router will use `cir.add_to_counter`, there would be different. Maybe not easy to extract to general logic.

src/ffi.rs
```diff
@@ -627,4 +820,74 @@ mod tests {
             assert!(errbuf_len < ERR_BUF_MAX_LEN);
         }
     }
+
+    #[cfg(feature = "expr_validation")]
+    #[test]
+    fn test_expression_validate() {
```
ProBrian, line 826: I think we'd better add test cases for `ATC_ROUTER_EXPRESSION_VALIDATE_FAILED` and `ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL`.

src/ffi.rs
Oyami-Srk: Given that `validate` is provided by a trait and `AST` should always implement this (I don't think we want to perform the validation on other data structures.) and `parse` is very natural from `AST` from `String`, I think `parse` and `validate` should be considered very stable interfaces. Indeed calling `parse` and then `validate` is a widespread use case but I don't think we should add a new interface just calling them together.

However, I did realize that we should not depend on `add_to_counter` for this interface thanks for your comment. Because `counter` should be an internal underlying method to achieve something that should not be public as the interface. Hai's PR changed the `semantics::FieldCounter` to be implemented by `Cir` rather than `AST` because of that "internality". This is an interface change and not a data structure change.

I'd like to introduce two traits like `semantics::GetFields` and `semantics::GetOperators` and implement this for `AST` only. This may provide a more idiomatic way to get data from `AST`.

WDYT? @dndx @ADD-SP @ProBrian

src/ffi.rs
Oyami-Srk: fixed.

src/ffi.rs
ProBrian: > I'd like to introduce two traits like `semantics::GetFields` and `semantics::GetOperators` and implement this for `AST` only. This may provide a more idiomatic way to get data from `AST`.

Sounds good, that will make the FFI layer more clear and brief.

src/ffi.rs
```diff
@@ -578,6 +578,199 @@ pub unsafe extern "C" fn context_get_result(
...
+    }
+
+    // Get used fields
+    if !(fields_buf.is_null() && fields_len.is_null() && fields_total.is_null()) {
+        if !fields_buf.is_null() {
+            assert!(
+                !fields_len.is_null() && !fields_total.is_null(),
+                "fields_len and fields_total must be non-null when fields_buf is non-null"
+            );
+        }
```
ADD-SP, line 697:
```suggestion
    if !fields_buf.is_null() {
        assert!(
          !fields_len.is_null() && ! fields_total.is_null(),
          "fields_len and fields_total must be non-null when fields_buf is non-null"
        )
```

Oyami-Srk: This would not allow a null `fields_buf` with non-null `fields_len` or `fields_total` to get the required buffer length and total field count without a valid buffer. This behavior is indeed NOT encouraged (since a waste of validating and getting fields/operators is called) but I think it's necessary to provide that form of interface to make it more flexible and keep consistency to other interfaces like `get_fields`.

src/semantics.rs
```diff
@@ -159,6 +174,74 @@ impl Validate for Expression {
...
+        predicates
+    }
+}
+
+#[cfg(feature = "expr_validation")]
+impl<T> GetFields for T
+where
+    T: GetPredicates,
+{
+    fn get_fields(&self) -> HashSet<String> {
```
ProBrian, line 217: Are there more forthcoming PRs? I see this function not been used in ffi.rs.
Is it possible that we just define this like `fn get_fields(predicates: &Vec<&Predicate>) -> HashSet<String>`, so that we can reuse the `predicates` list we got from `get_predicates` function, and no need to invoke `self.get_predicates()` again?

src/semantics.rs
```diff
@@ -159,6 +174,74 @@ impl Validate for Expression {
...
+        fields
+    }
+}
+
+#[cfg(feature = "expr_validation")]
+impl<T> GetOperators for T
+where
+    T: GetPredicates,
+{
+    fn get_operators(&self) -> crate::ast::BinaryOperatorFlags {
```
ProBrian, line 233: same as above comment, this function seems not been used.

src/semantics.rs
Oyami-Srk: I don't think `get_fields(predicates: &Vec<&Predicate>)` is a good interface since callers should not really care about what the `Predicate` is. Those two traits are indeed unused yet, but I'll make further changes to it.

Oyami-Srk: Done with commit 2cf3480, please check to see if this could be resolved.

ProBrian (approved):

src/ffi.rs
```diff
@@ -578,6 +578,180 @@ pub unsafe extern "C" fn context_get_result(
...
+///    and it must be properly aligned.
+/// - `errbuf_len` must be valid to read and write for `size_of::<usize>()` bytes,
+///    and it must be properly aligned.
+/// - If `fields_buf` is non-null, `fields_len` and `fields_total` must be non-null.
+/// - If `fields_buf` is null, `fields_len` and `fields_total` can be non-null
+///   for writing required buffer length and total number of fields.
+#[cfg(feature = "expr_validation")]
+#[no_mangle]
+pub unsafe extern "C" fn expression_validate(
+    atc: *const u8,
```
javierguerragiraldez, line 652: any reason to use `u8` for strings here, and `i8` elsewhere?

src/ffi.rs
```diff
@@ -578,6 +578,180 @@ pub unsafe extern "C" fn context_get_result(
...
+///
+/// Returns an integer value indicating the validation result:
+/// - ATC_ROUTER_EXPRESSION_VALIDATE_OK(0) if validation is passed.
+/// - ATC_ROUTER_EXPRESSION_VALIDATE_FAILED(1) if validation is failed.
+/// - ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL(2) if the provided fields buffer is not enough.
+///
+/// If `fields_buf` is null and `fields_len` or `fields_total` is non-null, it will write
+/// the required buffer length and the total number of fields to the provided pointers.
+/// If `fields_buf` is non-null, and `fields_len` is enough for the required buffer length,
+/// it will write the used fields to the buffer separated by '\0' and the total number of fields
```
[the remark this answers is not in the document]
Oyami-Srk: Sure, thanks for the suggestion. All fields got `'\0'` for the terminator.

src/ffi.rs
Oyami-Srk: According to the [document](https://doc.rust-lang.org/std/ffi/type.c_char.html), `ffi::c_char` is defined as `i8`. However, the document also mentioned that On modern architectures this type will always be either `i8` or `u8`.

Choosing `u8` here is just to keep consistency with Rust internal bytes represent like [`String::as_bytes`](https://doc.rust-lang.org/std/string/struct.String.html#method.as_bytes). 

P.S. I saw all the FFI interfaces in this project have no consistent way to represent `c_char`. E.g. `fields` is `u8` in `router_get_fields` and `i8` in `context_add_value`.

Oyami-Srk:
Rebased with the latest main branch. Nothing else has been changed except `Cargo.toml`, `Cargo.lock` and `src/semantics.rs` since 2 weeks ago.

src/ffi.rs
dndx: @Oyami-Srk In general, we tries to use `u8` for `char` whenever possible, except when dealing with char provided by LuaJIT.

This is because LuaJIT converts string to `const char *`, and this is a signed value. If we accept `u8` LuaJIT will complain when attempting to convert a string.

Oyami-Srk:
@javierguerragiraldez Could you please approve this if there's no other concern in this Rust implementation?

src/ffi.rs
```diff
@@ -578,6 +578,180 @@ pub unsafe extern "C" fn context_get_result(
...
+/// - `errbuf_len`: a pointer to the length of the error message buffer.
+///
+/// # Returns
+///
+/// Returns an integer value indicating the validation result:
+/// - ATC_ROUTER_EXPRESSION_VALIDATE_OK(0) if validation is passed.
+/// - ATC_ROUTER_EXPRESSION_VALIDATE_FAILED(1) if validation is failed.
+/// - ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL(2) if the provided fields buffer is not enough.
+///
+/// If `fields_buf` is null and `fields_len` or `fields_total` is non-null, it will write
```
dndx, line 608: I think we should just be optimistic here and always ask the user to pass in a buffer, 99% of the time reallocate will not be necessary, and we can avoid having to parse the expression twice which could be expensive (especially with Regexes).

dndx (changes requested):

Cargo.toml
```diff
@@ -32,6 +33,7 @@ crate-type = ["lib", "cdylib", "staticlib"]
 default = ["ffi"]
 ffi = []
 serde = ["cidr/serde", "dep:serde", "dep:serde_regex"]
+expr_validation = ["dep:bitflags"]
```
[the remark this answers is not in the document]
Oyami-Srk: Done. Now feature flag `expr_validation` has been removed.

src/ast.rs
```diff
@@ -45,6 +45,59 @@ pub enum BinaryOperator {
...
+            | Self::POSTFIX.bits()
+            | Self::GREATER.bits()
+            | Self::GREATER_OR_EQUAL.bits()
+            | Self::LESS.bits()
+            | Self::LESS_OR_EQUAL.bits()
+            | Self::IN.bits()
+            | Self::NOT_IN.bits()
+            | Self::CONTAINS.bits());
+    }
+}
```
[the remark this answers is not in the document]
Oyami-Srk: Done.

src/ast.rs
```diff
@@ -133,6 +186,17 @@ pub struct Predicate {
...
+impl Predicate {
+    pub fn get_field(&self) -> &str {
+        &self.lhs.var_name
+    }
+
+    pub fn get_operator(&self) -> &BinaryOperator {
+        &self.op
+    }
+}
+
```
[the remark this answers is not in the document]
Oyami-Srk: Done. Additionally, a private function to get predicates from `Expresison` has been moved under ffi module.

src/ffi.rs
```diff
@@ -578,6 +578,180 @@ pub unsafe extern "C" fn context_get_result(
...
 }
 
+#[cfg(feature = "expr_validation")]
+pub const ATC_ROUTER_EXPRESSION_VALIDATE_OK: i64 = 0;
+#[cfg(feature = "expr_validation")]
+pub const ATC_ROUTER_EXPRESSION_VALIDATE_FAILED: i64 = 1;
+#[cfg(feature = "expr_validation")]
+pub const ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL: i64 = 2;
+
+/// Validate the ATC expression with the schema.
```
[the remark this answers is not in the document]
Oyami-Srk: Done.

src/ffi.rs
Oyami-Srk: I've removed the ability to acquire the buffer length. Callers must pass a valid buffer if the calling is for getting the fields in use.
But I keep the `fields_buf` null-able because this interface is `expression_validate` and callers may not always want fields. This will allow us to do more refactoring on the Kong side with this interface. Like getting rid of validating expressions by inserting and then deleting them.

Oyami-Srk:
Did a little improvement:

1. Break `ffi.rs` into module.
2. Move everything guarded by the feature flag `expr_validation` to `ffi:expression`.
3. **Remove feature flag `expr_validation`**.
4. Remove public interfaces `get_field` and `get_operator` for `Predicate`.
5. Remove the redundant error message when the buffer is too small.
6. Remove the ability to fetch field count and total length without a valid `field_buf`. (The `field_buf` is still null-able to allow doing validation only.)

@dndx Could you review it when you have a moment?
@javierguerragiraldez Could you please re-generate the C header and update [cgo compile flag](https://github.com/Kong/go-atc-router/pull/6/files#diff-146f7e1ec8fd2395b53645c09467eddc8d3d5f9cb100235e546a36bedfb5a770R4) as the feature flag `expr_validation` is removed, and only need feature `ffi`.

javierguerragiraldez:
> @javierguerragiraldez Could you please re-generate the C header and update [cgo compile flag] as the feature flag `expr_validation` is removed, and only need feature `ffi`.

:+1:  done

src/ffi/expression.rs
```diff
@@ -0,0 +1,399 @@
...
+/// and the total number of fields to `fields_total`, then returns `ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL`.
+///
+/// If `operators` is non-null, it writes the used operators as bitflags to the provided pointer.
+/// Bitflags are defined by `BinaryOperatorFlags` and must exclude bits from `BinaryOperatorFlags::UNUSED`.
+///
+/// # Panics
+///
+/// This function will panic if:
+///
+/// - `atc` does not point to a valid C-style string.
```
dndx, line 132: This should be moved to "Safety" section, as "`atc` does not point to a valid C-style string" will be a safety violation and we can not guarantee a panic will happen.

src/ffi/expression.rs
```diff
@@ -0,0 +1,399 @@
...
+pub const ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL: i64 = 2;
+
+/// Validates an ATC expression against a schema.
+///
+/// # Arguments
+///
+/// - `atc`: a C-style string representing the ATC expression.
+/// - `schema`: a valid pointer to a [`Schema`] object, as returned by [`schema_new`].
+/// - `fields_buf`: a buffer for storing the fields used in the expression.
+/// - `fields_len`: a pointer to the length of `fields_buf`.
```
dndx, line 106:
```suggestion
/// - `fields_buf_len`: a pointer to the length of `fields_buf`.
```

I was a bit confused by the name, if this refers to the size of `fields_buf` we should make `fields_buf` part of the name as well.

src/ffi/expression.rs
```diff
@@ -0,0 +1,399 @@
...
+/// - `fields_total`: a pointer for storing the total number of fields.
```
dndx, line 107:
```suggestion
/// - `fields_total`: a pointer for storing the total number of unique fields used in the expression.
```

src/ffi/expression.rs
```diff
@@ -0,0 +1,399 @@
...
+/// An integer indicating the validation result:
+/// - `ATC_ROUTER_EXPRESSION_VALIDATE_OK` (0): Validation succeeded.
+/// - `ATC_ROUTER_EXPRESSION_VALIDATE_FAILED` (1): Validation failed; `errbuf` and `errbuf_len` will be updated with an error message.
+/// - `ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL` (2): The provided `fields_buf` is too small.
+///
+/// If `fields_buf` is non-null and `fields_len` is sufficient, this function writes the used fields to `fields_buf`,
+/// each field terminated by `\0`. It updates `fields_len` with the required buffer length and stores the total number of fields in `fields_total`.
+///
+/// If `fields_buf` is non-null but `fields_len` is insufficient, it writes the required buffer length to `fields_len`
+/// and the total number of fields to `fields_total`, then returns `ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL`.
```
dndx, line 123: Do we even care about `fields_total` when `ATC_ROUTER_EXPRESSION_VALIDATE_BUF_TOO_SMALL`? I'm not sure what can we do with this value if `fields_buf` is truncated.

src/ffi/expression.rs
```diff
@@ -0,0 +1,399 @@
...
+///
+/// If `operators` is non-null, it writes the used operators as bitflags to the provided pointer.
```
dndx, line 125: Do we have to make so much fields optional?

Doing so makes the documentation hard to understand and number of combinations explode, and not sure if it even provides us much value. Should we just make everything required always instead?

src/ffi/expression.rs
```diff
@@ -0,0 +1,399 @@
...
+use crate::ffi::ERR_BUF_MAX_LEN;
+use crate::schema::Schema;
+use bitflags::bitflags;
+use std::cmp::min;
+use std::ffi;
+use std::os::raw::c_char;
+use std::slice::from_raw_parts_mut;
+
+impl Expression {
+    fn get_predicates(&self) -> Vec<&Predicate> {
```
dndx, line 11: Not a hard blocker for merge, but should we consider using an `iter()` method instead? I noticed all of the use cases for returned `predicates` are iterations, and an iterator might be a better interface.

dndx (changes requested):

src/ffi/expression.rs
Oyami-Srk: Done.

src/ffi/expression.rs
Oyami-Srk: Done. Removed incorrect panic statement for `atc` due to the safety statements have alreay addressed `atc` pointer.

src/ffi/expression.rs
Oyami-Srk: The intend of optional fields `field_buf` and `operators` is to make this interface could validate expression without the overhead of getting elements from it. And hopefully be used by the [resty.router.router.validate](https://github.com/Kong/atc-router/blob/8c5d4f5dbdebead39c77e7cd7280e2a30eba4539/lib/resty/router/router.lua#L110).
I'm not updating `router.lua` because this PR is out of scope to do that.
But seemingly that `validate` method needs fields as well and it is not on the hot path. So it will be fine if we get `fields` and `operators` for every validating.

Oyami-Srk: I also removed the assertion on null since this is covered by Safety statement and will lead to UB if any of them are null.

src/ffi/expression.rs
Oyami-Srk: Currently, the implementation will not write any thing if buffer is too small. It will save all fields to be writen in a vector and always calcuate required length. This could be improved once the iterator is in use.

src/ffi/expression.rs
Oyami-Srk: Will do before this PR getting merged. Iteration interface will improve a little performance.

Oyami-Srk: Done. Now use a stack-based iterator. Although using Morris traversal could achieve this with more space efficiency (O(1) in space, O(n) in time), it is more intrusive on the tree structure itself and may require a raw pointer on the tree. I don't think it's worth it.

src/ffi/expression.rs
Oyami-Srk: Done.

ADD-SP (approved):
LGTM, just a small thing.

src/ffi/expression.rs
```diff
@@ -0,0 +1,382 @@
...
+    *fields_total = 0;
+
+    for pred in ast.iter_predicates() {
+        ops |= BinaryOperatorFlags::from(&pred.op);
+
+        let field = pred.lhs.var_name.as_str();
+
+        if existed_fields.insert(field) {
+            // Fields is not existed yet.
+            let field = ffi::CString::new(field).unwrap();
```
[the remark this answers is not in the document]
Oyami-Srk: Done.

src/ffi/context.rs
```diff
@@ -0,0 +1,251 @@
...
+use std::ffi;
+use std::os::raw::c_char;
+use std::slice::from_raw_parts_mut;
+use uuid::fmt::Hyphenated;
+
+/// Allocate a new context object associated with the schema.
+///
+/// # Errors
+///
+/// This function never fails.
```
dndx, line 15:
```suggestion
/// This function never returns an error, however, it can panic if memory allocation failed.
```

dndx (approved):

PR #279, opened 2024-11-20
chore(*): remove Bazel and bump dependencies, release `v1.7.0`

dndx:
We don't need the Bazel files inside this project anymore as it is not done inside Kong's build system directly. This also means we no longer need to store `Cargo.lock` inside this project.


ADD-SP (commented):
We can also delete the `.github/workflows/bazel.yml` to make the CI happy.

ADD-SP (approved):

Oyami-Srk (approved):

PR #286, opened 2024-12-30
perf(cir): special treatment for single-predicate expression

Oyami-Srk:
[KAG-6094](https://konghq.atlassian.net/browse/KAG-6094)

[KAG-6094]: https://konghq.atlassian.net/browse/KAG-6094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cir.rs
+++ b/src/cir.rs
@@ -6,22 +6,9 @@ use crate::semantics::FieldCounter;
 use std::collections::HashMap;
 
 #[derive(Debug)]
-pub struct CirProgram {
-    pub(crate) instructions: Vec<CirInstruction>,
-}
-
-impl CirProgram {
-    pub fn new() -> Self {
-        Self {
-            instructions: Vec::new(),
-        }
-    }
-}
-
-impl Default for CirProgram {
-    fn default() -> Self {
-        Self::new()
-    }
+pub enum CirProgram {
+    Instructions(Vec<CirInstruction>),
+    Predicate(Predicate),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/cir.rs
```diff
@@ -6,22 +6,9 @@ use crate::semantics::FieldCounter;
...
-        }
-    }
-}
-
-impl Default for CirProgram {
-    fn default() -> Self {
-        Self::new()
-    }
+pub enum CirProgram {
+    Instructions(Vec<CirInstruction>),
```
Oyami-Srk, line 10: Should we use boxed slice instead `Vec` to emphasize that instructions won't grow over time?

xianghai2 (approved):

src/cir.rs
Oyami-Srk: Fixed. It's not hurt performance, though. The only difference between `Vec` and `Box<[]>` is the `capacity` field.

src/cir.rs
```diff
@@ -63,10 +56,27 @@ pub trait Translate {
 impl Translate for Expression {
     type Output = CirProgram;
     fn translate(&self) -> Self::Output {
-        let mut cir = CirProgram::new();
-        cir_translate_helper(self, &mut cir);
-        cir.instructions.shrink_to_fit(); // shrink the memory
-        cir
+        let mut cir_instructions = Vec::new();
+        cir_translate_helper(self, &mut cir_instructions);
+        cir_instructions.shrink_to_fit();
```
ADD-SP, line 61:
```suggestion
```

The `Vec::into_boxed_slice()` also discards excess capacity.

Oyami-Srk: committed.

PR #292, opened 2025-03-24
feat(router): add `try_match` function

michaelbeaumont:
- Needs https://github.com/Kong/atc-router/pull/291

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -55,17 +55,12 @@ impl<'a> Router<'a> {
     }
 
     pub fn execute(&self, context: &mut Context) -> bool {
-        for (MatcherKey(_, id), m) in self.matchers.iter().rev() {
-            let mut mat = Match::new();
-            if m.execute(context, &mut mat) {
-                mat.uuid = *id;
-                context.result = Some(mat);
-
-                return true;
-            }
+        if let Some(m) = self.r#match(context) {
+            context.result = Some(m);
+            true
+        } else {
+            false
         }
-
-        false
     }
 
     /// Note that unlike `execute`, this doesn't set `Context.result`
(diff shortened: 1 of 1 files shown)


src/router.rs
```diff
@@ -66,4 +66,18 @@ impl<'a> Router<'a> {
 
         false
     }
+
+    /// Note that unlike `execute`, this doesn't set `Context.result`
+    /// but it also doesn't need a `&mut Context`.
+    pub fn r#match(&self, context: &Context) -> Option<Match> {
```
Oyami-Srk, line 72: Could you please write a test for this?

Oyami-Srk, line 72: Can we have another name rather than `match`? The raw identifier is too verbose to be used.

Oyami-Srk: Like `find` or `search`?

Oyami-Srk:
Shall we also export this function to FFI?

src/router.rs
michaelbeaumont: Actually what about `try_match`?

michaelbeaumont: find/search sound too much like regex/globs and here we're executing "matchers"

michaelbeaumont:
> Shall we also export this function to FFI? Don't think it's needed, but is there any requirement for keeping a consistent interface?

I suppose not exporting something for now is safer? We're using this directly as a rust crate so it's not important for us.

src/router.rs
michaelbeaumont: done, also added some for `execute`

Oyami-Srk (approved):
LGTM, Thanks. We can add FFI interface once we really need it.

ADD-SP (approved):

PR #299, opened 2025-03-29
feat: use ahash

### LINKED ISSUE #298: Use `ahash` instead of `fnv` for `HashMap`
The [current implementation](https://github.com/Kong/atc-router/blob/ffd11db657115769bf94f0c4f915f98300bc26b6/src/context.rs#L6-L10) of atc-router uses FnvHash,

https://github.com/Kong/atc-router/blob/ffd11db657115769bf94f0c4f915f98300bc26b6/src/context.rs#L6-L10

which, according to the [ahash documentation](https://github.com/tkaitchuck/aHash/tree/master/compare#fnvhash):

> FnvHash is also a poor default. It only handles one byte at a time, so its performance really suffers with > large inputs. It is also non-keyed so it is still subject to DOS attacks and [accidentally quadratic behavior.](https://accidentallyquadratic.tumblr.com/post/153545455987/rust-hash-iteration-reinsertion)

is inefficient with large inputs due to byte-by-byte processing and vulnerable to DOS attacks and [accidentally quadratic behavior](https://accidentallyquadratic.tumblr.com/post/153545455987/rust-hash-iteration-reinsertion) because it’s non-keyed.

In contrast, [AHash](https://github.com/tkaitchuck/ahash) offers superior performance, especially on x86_64 with SIMD AES-NI instructions, and enhanced security across all platforms, with reliable fallbacks elsewhere.

Colerar:
Closes #298

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,8 @@ uuid = "1.8"
 regex = "1"
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_regex = { version = "1.1", optional = true }
-fnv = "1"
 bitflags = { version = "2.6", optional = true }
+ahash = "0.8"
 
 [dev-dependencies]
 criterion = "0"
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,21 +1,21 @@
 use crate::ast::Value;
 use crate::schema::Schema;
-use fnv::FnvHashMap;
+use ahash::AHashMap;
 use uuid::Uuid;
 
 #[derive(Debug)]
 pub struct Match {
     pub uuid: Uuid,
-    pub matches: FnvHashMap<String, Value>,
-    pub captures: FnvHashMap<String, String>,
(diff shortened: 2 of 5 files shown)


dndx:
Hello @Colerar , thanks for your contribution!

Regarding the hashing algorithm, note the key used inside the `HashMap` here can *not* be controlled by the incoming request, and can only be set by the admin, and the length are generally very short (< 16 characters), therefore using ahash probably wouldn't generate meaningful security or performance improvements.

Colerar:
Yes, thanks for the context! `AHash` is do not suitable here.

But fnv is a somewhat old hasher and we may have better options. We could benchmark possible alternatives:

- [foldhash](https://github.com/orlp/foldhash): the default hasher of `hashbrown`
- [rustc-hash](https://github.com/rust-lang/rustc-hash): used by rustc. (It’s worth mentioning that rustc also used fnv before)

PR #302, opened 2025-04-13
style(src): fix some code style issues

chronolaw:
KAG-6866
wait for #305

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,21 +2,21 @@ use crate::ast::Value;
 use crate::schema::Schema;
 use uuid::Uuid;
 
-type MyHashMap<K, V> = fnv::FnvHashMap<K, V>;
+type HashMap<K, V> = fnv::FnvHashMap<K, V>;
 
 #[derive(Debug)]
 pub struct Match {
     pub uuid: Uuid,
-    pub matches: MyHashMap<String, Value>,
-    pub captures: MyHashMap<String, String>,
+    pub matches: HashMap<String, Value>,
+    pub captures: HashMap<String, String>,
 }
 
 impl Match {
     pub fn new() -> Self {
         Match {
             uuid: Uuid::default(),
-            matches: MyHashMap::default(),
-            captures: MyHashMap::default(),
+            matches: HashMap::default(),
+            captures: HashMap::default(),
         }
     }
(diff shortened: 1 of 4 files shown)


chobits (approved):

src/context.rs
```diff
@@ -1,21 +1,22 @@
 use crate::ast::Value;
 use crate::schema::Schema;
-use fnv::FnvHashMap;
 use uuid::Uuid;
 
+type HashMap<K, V> = fnv::FnvHashMap<K, V>;
```
Oyami-Srk, line 5: Might be confused with `std::collections::HashMap`

chronolaw: Yes, but I can not think about a new name? do you have any suggestion?

chronolaw: rename to FastHashMap

Oyami-Srk (commented):
Actually, I don't get the necessity of this change. WDYT @ADD-SP

chronolaw:
> Actually, I don't get the necessity of this change. WDYT @ADD-SP

It lets us change the hashmap easily, like `type FastHashMap = another implementation`

src/context.rs
```diff
@@ -1,21 +1,22 @@
 use crate::ast::Value;
 use crate::schema::Schema;
-use fnv::FnvHashMap;
 use uuid::Uuid;
 
+type FastHashMap<K, V> = fnv::FnvHashMap<K, V>;
```
ADD-SP, line 5: This may not be the common pattern in Rust.

In Rust, something like `Pin<Box<dyn Future<Output = Result<T, Box<dyn Error + Send>>> + Send` might have an alias because of the crazy generic bounds.

ADD-SP:
> > Actually, I don't get the necessity of this change. WDYT @ADD-SP
> 
> It lets us change the hashmap easily, like `type FastHashMap = another implementation`

It is unlikely for us to change the hashmap implementation frequently.

chronolaw:
OK, I will revert the changes of hash map, only keep the style fixes.

src/context.rs
chronolaw: reverted.

Oyami-Srk (approved):

PR #303, opened 2025-04-14
refactor(src): simplify `if let` return statement

chronolaw:
Same style as `remove_matcher()`
KAG-6866

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -238,6 +238,7 @@ fn parse_predicate(pair: Pair<Rule>) -> ParseResult<Predicate> {
         op,
     })
 }
+
 // transform_func = { ident ~ "(" ~ lhs ~ ")" }
 #[allow(clippy::result_large_err)] // it's fine as parsing is not the hot path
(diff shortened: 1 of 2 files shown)


chronolaw:
lint will be fixed in #302.

ADD-SP (commented):
How about this?

``` rust
        let Some(m) = self.try_match(context) else {
            return false;
        };
        context.result = Some(m);
        true
```

chronolaw:
> How about this?
> 
> ```rust
>         let Some(m) = self.try_match(context) else {
>             return false;
>         };
>         context.result = Some(m);
>         true
> ```

Comparing to original code it dose not seem to be more clean.

Oyami-Srk:
I think an early return of false will make future changes to the "matched" path clearer.

chronolaw:
> I think an early return of false will make future changes to the "matched" path clearer.

OK, changed.

Oyami-Srk (approved):

ADD-SP (approved):

PR #304, opened 2025-04-14
refactor(src): simplify code with macro

chronolaw:
KAG-6866

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -76,8 +76,8 @@ impl Execute for Predicate {
                     }
                 }
                 BinaryOperator::Regex => {
-                    let rhs = get_op_value!(&self.rhs, Value::Regex);
                     let lhs = get_op_value!(lhs_value, Value::String);
+                    let rhs = get_op_value!(&self.rhs, Value::Regex);
 
                     if rhs.is_match(lhs) {
                         let reg_cap = rhs.captures(lhs).unwrap();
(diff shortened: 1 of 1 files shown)


ADD-SP (commented):
The better practice in Rust to improve this case is implement `as_*() -> Option<>` for the enum value.

For example: https://docs.rs/serde_json/latest/serde_json/enum.Value.html

Could you try this?

chronolaw:
Currently the code has `unreachable!`, it may not be implemented to return an Option.
I could try it after this PR is merged.

ADD-SP:
> Currently the code has `unreachable!`, it may not be implemented to return an Option. I could try it after this PR is merged.

We don't need to return the `Option`, we just need to panic once we got a `None`.

```
let rhs = self.rhs.as_string().expect("should be a string");
```

chronolaw:
> 
> We don't need to return the `Option`, we just need to panic once we got a `None`.
> 
> ```rust
> let rhs = self.rhs.as_str().expect("should be a string unless we generated incorrect AST");
> ```

This changed the current logic, I suggest to implement it in the next PR.

ADD-SP:
> > We don't need to return the `Option`, we just need to panic once we got a `None`.
> > ```rust
> > let rhs = self.rhs.as_str().expect("should be a string unless we generated incorrect AST");
> > ```
> 
> This changed the current logic, I suggest to implement it in the next PR.

Why does this change the current logic? Maybe I missed something? Calling the `.expect` on a `None` also panics the program, just like `unreachable!`.

chronolaw:
Got, I will try it in another PR.

chronolaw:
replaced by #306.

PR #306, opened 2025-05-06
refactor(src): simplify code with `as_xxx`

chronolaw:
Replacement of #304
KAG-6866

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -84,9 +84,9 @@ impl Value {
 }
 
 impl Value {
-    pub fn as_string(&self) -> Option<&String> {
+    pub fn as_string(&self) -> Option<&str> {
         if let Value::String(s) = self {
-            return Some(s);
+            return Some(s.as_str());
         }
         None
     }
@@ -98,9 +98,9 @@ impl Value {
         None
     }
 
-    pub fn as_int(&self) -> Option<&i64> {
+    pub fn as_int(&self) -> Option<i64> {
         if let Value::Int(i) = self {
-            return Some(i);
+            return Some(*i);
         }
         None
     }
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -83,6 +83,8 @@ impl Value {
     }
 }
 
+// as_xxx().unwrap is safe and None is unreachable
+// when we call it as its correct value
 impl Value {
     pub fn as_str(&self) -> Option<&str> {
         if let Value::String(s) = self {
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -111,6 +111,13 @@ impl Value {
         }
         None
     }
+
+    pub fn as_ipcidr(&self) -> Option<&IpCidr> {
+        if let Value::IpCidr(c) = self {
+            return Some(c);
+        }
+        None
+    }
 }
 
 impl From<String> for Value {
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -83,8 +83,6 @@ impl Value {
     }
 }
 
-// as_xxx().unwrap is safe and None is unreachable
-// when we call it as its correct value
 impl Value {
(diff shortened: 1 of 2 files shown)


src/ast.rs
```diff
@@ -81,6 +81,27 @@ impl Value {
             Value::Regex(_) => Type::Regex,
         }
     }
+
+    pub fn as_string(&self) -> Option<&String> {
```
ADD-SP, line 87: Rust always suggests `&str` instead of `&String`.

src/ast.rs
```diff
@@ -81,6 +81,27 @@ impl Value {
...
+        if let Value::String(s) = self {
+            return Some(s);
+        }
+        None
+    }
+
+    pub fn as_int(&self) -> Option<&i64> {
```
ADD-SP, line 101: For copyable type like `i64`, we'd better doing copy-and-return.

chronolaw: done.

chobits (approved):

src/ast.rs
chobits: Rust often suggests using &str (a string slice) instead of &String (a reference to a String) in function signatures and other places where you're borrowing string data

&str is more general: A &str can point to either a String (which owns its data on the heap) or a string literal (which is typically stored in the program's binary).

&String is more specific: A &String can only refer to a String.


By accepting &str, your function becomes more versatile. It can work with both types of string data without requiring the caller to explicitly convert a string literal to a String.

chobits: More versatile for your code, if you use `str`

Oyami-Srk (approved):
nit: according to the code standard, it's better to explain that the `unwrap` is unreachable in comments.

ADD-SP:
> > nit: according to the code standard, it's better to explain that the `unwrap` is unreachable in comments.
> 
> Could you give out an example or the code standard item? thanks.

This is the recommended style in Rust when calling a function that may panic. If the correctness is not obvious, developer may add safety comment to explain why this is correct.

For example

```
// This never panics unless the semantic checking doesn't catch
// the mismatched type, which should be a bug.
let s = lhs_value.as_str().unwrap();
```

Oyami-Srk:
@ADD-SP `IpAddr` and `IpCidr` are `Copy`. We can just copy them, but I think there's no benefit yet compared with the current usage of the references.

src/ast.rs
```diff
@@ -83,6 +83,45 @@ impl Value {
     }
 }
 
+// as_xxx().unwrap is safe and None is unreachable
+// when we call it as its correct value
```
ADD-SP, line 87: Safety should be nearby the code that may panics.

src/interpreter.rs
```diff
@@ -67,14 +66,11 @@ impl Execute for Predicate {
...
-                    };
-                    let lhs = match lhs_value {
-                        Value::String(s) => s,
-                        _ => unreachable!(),
-                    };
+                    debug_assert!(lhs_value.my_type() == Type::String);
+                    debug_assert!(self.rhs.my_type() == Type::Regex);
+
+                    let lhs = lhs_value.as_str().unwrap();
+                    let rhs = self.rhs.as_regex().unwrap();
```
ADD-SP, line 73:
```suggestion
                    // SAFETY: this only panic if and only if 
                    // the semantic checking didn't
                    // catch the mismatched types,
                    // which is a bug.
                    let lhs = lhs_value.as_str().unwrap();
                    let rhs = self.rhs.as_regex().unwrap();
```

I think a safety comment is enough to explain this case.

chronolaw: added.

src/ast.rs
chronolaw: removed.

ADD-SP (approved):

PR #309, opened 2025-06-19
refactor(src): clean for interpreter.rs

chronolaw:


chronolaw:
Clippy should be fixed in #311.

PR #310, opened 2025-06-19
tests(src): unit tests for schema

chronolaw:

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -18,3 +18,25 @@ impl Schema {
         self.fields.insert(field.to_string(), typ);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


src/schema.rs
```diff
@@ -18,3 +18,40 @@ impl Schema {
...
+    fn normal_fields() {
+        let mut s = Schema::default();
+
+        s.add_field("str", Type::String);
+        s.add_field("ip", Type::IpAddr);
+        s.add_field("cidr", Type::IpCidr);
+        s.add_field("r", Type::Regex);
+        s.add_field("i", Type::Int);
+
+        assert_eq!(s.type_of("str"), Some(&Type::String));
```
ADD-SP, line 36: nit: it should be ok to implement `Copy` to `Type`, so that we don't need `&` here.

PR #311, opened 2025-06-19
refactor(src): add `reset()` for Match

chronolaw:

It may improve the performance of `try_match()` a little.
In `m.execute()`, `mat` may allocate memory in hashmap, the PR could reuse these memory to avoid re-allocating.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -18,6 +18,11 @@ impl Match {
             captures: FnvHashMap::default(),
         }
     }
+
+    pub fn reset(&mut self) {
+        self.matches.clear();
+        self.captures.clear();
+    }
 }
 
 impl Default for Match {
--- a/src/context.rs
+++ b/src/context.rs
@@ -19,7 +19,7 @@ impl Match {
         }
     }
 
-    pub fn reset(&mut self) {
+    pub(crate) fn reset(&mut self) {
         self.matches.clear();
         self.captures.clear();
     }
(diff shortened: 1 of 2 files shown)


Oyami-Srk (approved):
IIRC, this optimization will be captured by existing (unmerged) benchmarks? If not, I think it's good to get some, not a blocker though.

ProBrian (approved):

chronolaw:
Actually it is a refactor, the performance improvement is not promised.

chronolaw:
Compare to branch "bench/match_mix":
```
   Compiling atc-router v1.7.1 (/home/chrono/rust/atc-router)
    Finished `bench` profile [optimized] target(s) in 3.18s
     Running benches/match_mix.rs (/home/chrono/rust/atc-router/target/release/deps/match_mix-d409b144ed8f34aa)
Gnuplot not found, using plotters backend
Match                   time:   [75.624 ns 75.731 ns 75.864 ns]
                        change: [-2.4785% -2.0091% -1.6081%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  4 (4.00%) high mild
  9 (9.00%) high severe

Doesn't Match           time:   [33.732 ms 33.789 ms 33.852 ms]
                        change: [-3.5870% -3.3017% -3.0365%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe
```

PR #313, opened 2025-06-24
feat: support adding Expressions as matchers to Routers

michaelbeaumont:


michaelbeaumont:
Because we may have already parsed Expressions that we'd like to add to the Router.

Oyami-Srk (approved):

michaelbeaumont:
@Oyami-Srk @chronolaw could we merge this now?

PR #315, opened 2025-06-25
feat: derive `PartialEq`, `Eq` for `ast::Expression`

michaelbeaumont:
Should we also `impl`/`derive` `Eq`?

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -7,14 +7,14 @@ use std::net::IpAddr;
 use serde::{Deserialize, Serialize};
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


ADD-SP:
> > In short, if AST is symmetric, then a == b if and only if b == a, however, clearly, AST is not symmetric.
> 
> Hmm why isn't it symmetric? What would be a counterexample?

@michaelbeaumont 

The most tricky thing is what is the definition of AST `a` is equals to AST `b`.

## Structurally Identical and node values are equal

For this definition, AST is symmetric.

## Semantically Identical

For this definition, AST is not symmetric.

ADD-SP:
> > In short, if AST is symmetric, then a == b if and only if b == a, however, clearly, AST is not symmetric.
> 
> Hmm why isn't it symmetric? What would be a counterexample?

@michaelbeaumont Ah, my fault, AST is symmetric.

PR #318, opened 2025-07-08
feat(*): schema owned or referenced

samugi:
This PR introduces a new way of initializing ATC so that it *owns* the schema that it depends on, instead of relying on a shared reference.

This facilitates the workflow especially when ATC is instantiated and used in Rust libraries, particularly in use cases where it needs to be cached for reuse - so that it is not necessary to maintain a reference to an external Schema object.

This is based on @lucab 's [work](https://github.com/lucab/atc-router/pull/1/files) and includes some small changes that are a follow up to his observations (same PR).

HOUD-98

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -5,6 +5,7 @@ use crate::parser::parse;
 use crate::schema::Schema;
 use crate::semantics::{FieldCounter, Validate};
 use std::collections::{BTreeMap, HashMap};
+use std::ops::Deref;
 use uuid::Uuid;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
@@ -12,18 +13,44 @@ struct MatcherKey(usize, Uuid);
 
 #[derive(Debug)]
 pub struct Router<'a> {
-    schema: &'a Schema,
+    schema: SchemaOwnedOrRef<'a>,
     matchers: BTreeMap<MatcherKey, Expression>,
     pub fields: HashMap<String, usize>,
 }
 
 impl<'a> Router<'a> {
+    /// Creates a new [`Router`] that holds a shared reference to a [`Schema`].
+    ///
+    /// This is useful when the schema is managed outside the router and/or shared
+    /// across multiple components.
     pub fn new(schema: &'a Schema) -> Self {
         Self {
-            schema,
+            schema: SchemaOwnedOrRef::Ref(schema),
             matchers: BTreeMap::new(),
             fields: HashMap::new(),
         }
+    }
+
+    /// Creates a new [`Router`] that owns its [`Schema`].
+    ///
+    /// This allows the router to be self contained,
+    /// making it easier to use as a standalone component.
+    pub fn new_owning(schema: Schema) -> Self {
+        Self {
+            schema: SchemaOwnedOrRef::Owned(schema),
+            matchers: BTreeMap::new(),
+            fields: HashMap::new(),
+        }
+    }
+
+    /// Returns a reference to the [`Schema`] used by this router.
+    ///
+    /// Especially useful if the router owns the schema internally ([`new_owning`]),
+    /// but you still need to pass a reference to other components like [`Context`].
+    ///
+    /// [`new_owning`]: Router::new_owning
+    pub fn schema(&self) -> &Schema {
+        &self.schema
     }
 
     pub fn add_matcher(&mut self, priority: usize, uuid: Uuid, atc: &str) -> Result<(), String> {
@@ -44,7 +71,7 @@ impl<'a> Router<'a> {
             return Err("UUID already exists".to_string());
         }
 
-        expr.validate(self.schema)?;
+        expr.validate(&self.schema)?;
         expr.add_to_counter(&mut self.fields);
 
         assert!(self.matchers.insert(key, expr).is_none());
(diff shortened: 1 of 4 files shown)


src/router.rs
```diff
@@ -5,27 +5,57 @@ use crate::parser::parse;
...
+use std::ops::Deref;
 use uuid::Uuid;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 struct MatcherKey(usize, Uuid);
 
+#[derive(Debug)]
+enum SchemaOwnedOrRef<'a> {
+    Ref(&'a Schema),
+    Owned(Box<Schema>),
```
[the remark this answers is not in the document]
Oyami-Srk: Originally, we access schema via a reference, so `Box` here doesn't make it worse, I think. If cache creation is going to be on the hot path, we should be careful about the CPU cache; both a bigger structure and indirect access could increase the cache miss rate. We found a regression caused by cache miss early, see https://github.com/Kong/atc-router/pull/286. However, I can't tell whether this is caused by the increment of the size of the struct or indirect access.

src/router.rs
```diff
@@ -5,27 +5,57 @@ use crate::parser::parse;
...
     pub fn new(schema: &'a Schema) -> Self {
         Self {
-            schema,
+            schema: SchemaOwnedOrRef::Ref(schema),
+            matchers: BTreeMap::new(),
+            fields: HashMap::new(),
+        }
+    }
+
+    pub fn new_owning(schema: Schema) -> Self {
```
Oyami-Srk, line 47: NIT: might be a little bit weird using `owning` here? I might be wrong.
```suggestion
    pub fn new_owned(schema: Schema) -> Self {
```

samugi: new "owning" was meant as in: this constructor generates a new "owning" router, i.e. one that owns a Schema

src/router.rs
```diff
@@ -5,27 +5,57 @@ use crate::parser::parse;
...
 use crate::semantics::{FieldCounter, Validate};
 use std::collections::{BTreeMap, HashMap};
+use std::ops::Deref;
 use uuid::Uuid;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 struct MatcherKey(usize, Uuid);
 
+#[derive(Debug)]
+enum SchemaOwnedOrRef<'a> {
```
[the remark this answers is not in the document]
samugi: done, thanks.

src/router.rs
```diff
@@ -5,27 +5,78 @@ use crate::parser::parse;
...
+///
+/// Used by [`Router`] to support both externally managed and self-contained schemas.
+/// Owning the schema is especially useful when the router is used outside of the FFI context,
+/// making it fully independent.
+///
+/// Implements [`Deref`] for ergonomic access to the underlying [`Schema`].
+#[derive(Debug)]
+enum SchemaOwnedOrRef<'a> {
+    Ref(&'a Schema),
+    Owned(Box<Schema>),
```
samugi, line 31: an alternative here could be using an Arc so that the same `Arc<Schema>` can be shared across routers while using the `Owned` variant:
```suggestion
    Owned(Arc<Schema>),
```

---

I also considered using an `Arc` **instead** of the current approach (to replace the `SchemaOwnedOrRef` pointer entirely), but it seems this would require modifying the FFI interface, which is probably a larger scope than we want for this PR.

Looking for some feedback on this @dndx - thanks

src/router.rs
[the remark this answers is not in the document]
Oyami-Srk: I'm concerned that we can't tell CPU cache level regression in a unit benchmark because it has plenty of resources. However, for Kong's usage, slowing down context creation slightly doesn't harm the proxy path, so I'm okay with both directions. We can use an unboxed variant for that and optimize this once it becomes a real problem.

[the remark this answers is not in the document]
samugi: I changed it back to unboxed since that appears to be the preferred approach

src/router.rs
samugi: For example, `router_new` [here](https://github.com/Kong/atc-router/blob/afe0b973eb971e0b9dd34d99f05ec1cd3ff27353/src/ffi/router.rs#L27):

> pub unsafe extern "C" fn router_new(schema: &Schema) -> *mut Router 

If we were to change our Router struct to something like:
```rust
pub struct Router {
    schema: Arc<Schema>,
    matchers: BTreeMap<MatcherKey, Expression>,
    pub fields: HashMap<String, usize>,
}
```

Then in order to maintain the ability to share a schema across routers we should have:

```rust
pub fn new(schema: Arc<Schema>) -> Self {
    Self {
        schema,
        matchers: BTreeMap::new(),
        fields: HashMap::new(),
    }
}
```

but that would be incompatible with the current schema parameter passed to `router_new` as a reference (which I think we want to maintain unchanged, at least for the scope of this PR).

Oyami-Srk: Oh, I thought we were just changing `enum SchemaOwnedOrRef` to use `Arc` instead of `Box`, and that will not affect the previous reference-based interface. For changing `Router.schema`, it indeed introduces a breaking change to our FFI, I'm not sure whether this is feasible.

samugi: yes, sorry, let me rephrase that.

What I meant is that the current change (using SchemaOwnedOrRef) is safe, changing the owned Schema to use Arc or Box is also safe and we could decide to do that easily (but based on another [closed discussion](https://github.com/Kong/atc-router/pull/318#discussion_r2192414371) above we are probably not going to anyway).

An alternative approach where, instead of using `SchemaOwnedOrRef`, we just replace the current Schema reference with an `Arc` would require an FFI interface change as explained above, and I agree with you that it is probably not feasible, so I think we are on the same page.

Oyami-Srk (approved):
LGTM. Hope there is no performance regression caused by 40 bytes larger `Router` in the next bump of atc-router.

lucab (approved):

PR #295, opened 2025-03-26
Reapply "perf(interpreter): improve ATC interprater's cache efficiency (#246)"

Oyami-Srk:

PR #322, opened 2025-07-15
perf(src): introduce cir for expression

chronolaw:
Related PR: #295 #296
https://konghq.atlassian.net/browse/KAG-7369


Oyami-Srk:
The performance report of Kong with this PR indicates that there is no significant performance improvement. See https://konghq.atlassian.net/browse/KAG-7386

PR #323, opened 2025-07-15
refactor(src): reuse types in semantics.rs

chronolaw:


src/semantics.rs
```diff
@@ -2,8 +2,8 @@ use crate::ast::{BinaryOperator, Expression, LogicalExpression, Type, Value};
 use crate::schema::Schema;
 use std::collections::HashMap;
 
-type ValidationResult = Result<(), String>;
-type ValidationHashMap = HashMap<String, usize>;
+pub type ValidationResult = Result<(), String>;
```
ProBrian, line 5: I think this is a good behavior like `using` or `typedef` in cpp.
We should confirm what's our preferred practice in type definition behavior in Rust, maybe we  could have suggestion added in the Rust coding standard.

ProBrian (approved):

src/semantics.rs
```diff
@@ -2,8 +2,8 @@ use crate::ast::{BinaryOperator, Expression, LogicalExpression, Type, Value};
...
+pub type ValidationHashMap = HashMap<String, usize>;
```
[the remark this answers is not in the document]
chronolaw: Yes, It is a good idea, I think that we could do it in the next refactor PR.

src/semantics.rs
jieyouxu: Remark (drive-by comment): To add, a closely-related technique is the [*Newtype* Idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html). When to use newtypes versus type aliases typically come down to properties you wish to enforce/realize:

- Newtypes are typically used when you have the need for *strong* guarantees that you cannot mix two values of the same underlying type. Just because two values are of the same *underlying* type, doesn't mean you can mix them together. The easiest example for this are various ID newtypes. For instance, if you have two arenas both storing `u32`s, you would want to use newtypes to enforce that you do not mix IDs into the arenas. The tradeoff is usually more implementation complexity (esp. if you have to forward some methods).
    - An additional benefit is that you can layer a more semantically meaningful API on top of the underlying type. 
- Type aliases are nice when it's *acceptable* to mix values of the same underlying type, and is used as a convenience "shorthand", or a "more meaningful" aliased type name.

PR #324, opened 2025-07-16
chore(*): Router takes Borrow<Schema> parameter

samugi:
update the router to accept types that implement `Borrow<Schema>` instead of using the `SchemaOwnedOrRef` type and multiple constructors.

This allows for more flexibility in how schemas are passed to the router while keeping the size of the router unchanged for FFI instantiations.

[HOUD-98]

[HOUD-98]: https://konghq.atlassian.net/browse/HOUD-98?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -4,40 +4,30 @@ use crate::interpreter::Execute;
 use crate::parser::parse;
 use crate::schema::Schema;
 use crate::semantics::{FieldCounter, Validate};
+use std::borrow::Borrow;
 use std::collections::{BTreeMap, HashMap};
-use std::ops::Deref;
 use uuid::Uuid;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 struct MatcherKey(usize, Uuid);
 
 #[derive(Debug)]
-pub struct Router<'a> {
-    schema: SchemaOwnedOrRef<'a>,
+pub struct Router<S> {
+    schema: S,
     matchers: BTreeMap<MatcherKey, Expression>,
     pub fields: HashMap<String, usize>,
 }
 
-impl<'a> Router<'a> {
-    /// Creates a new [`Router`] that holds a shared reference to a [`Schema`].
+impl<S> Router<S>
+where
+    S: Borrow<Schema>,
+{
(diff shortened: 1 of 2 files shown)


samugi:
Not sure if this is considered a breaking change: the router type lost a lifetime and gained a parameter.

src/router.rs
```diff
@@ -5,52 +5,39 @@ use crate::parser::parse;
...
-use std::ops::Deref;
 use uuid::Uuid;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 struct MatcherKey(usize, Uuid);
 
 #[derive(Debug)]
-pub struct Router<'a> {
-    schema: SchemaOwnedOrRef<'a>,
+pub struct Router<S> {
```
Oyami-Srk, line 14: I'd like to suggest adding a bound here. As `Router<S>` without `AsRef<Schema>` doesn't make any sense here.

```suggestion
pub struct Router<S>
where
    S: AsRef<Schema>
{
```

samugi: thanks for the suggestion @Oyami-Srk - I had this exact conversation with @lucab who suggested to **remove the bound** because having it would come with a list of downsides that are explained very well in [this SO answer](https://stackoverflow.com/a/66369912/1619736). I don't feel too strongly about going either direction but the points there make the no-bound option feel more appropriate to me. What do you think?

Oyami-Srk: Got it! Thanks for the clarification.

dndx: I think we should leave the bound out of the type definition, which is the way Rust standard library does it as well.

src/schema.rs
```diff
@@ -19,6 +19,12 @@ impl Schema {
     }
 }
 
+impl AsRef<Schema> for Schema {
+    fn as_ref(&self) -> &Schema {
+        self
+    }
+}
+
```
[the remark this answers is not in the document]
samugi: cool. I'll update the PR, thank you both

dndx:
@Oyami-Srk Could you take a look and ✅ if it looks good to you as well?

PR #328, opened 2025-08-13
feat: bool value

### LINKED ISSUE #325: boolean type
Right now I see
```
pub enum Type {
    String,
    IpCidr,
    IpAddr,
    Int,
    Regex,
}
```
I'd like to add a Boolean field to my schema.

Use case: I'm adding field to context called `kafka.record.value.validated` to indicate whether the Kafka record was validated. Based on this information we want to execute specific policies.

jakubdyszkiewicz:
This PR adds a bool value type to ATC. It's not clear to me what to do with `lib/resty/router` files. It says it's autogenerated, but there are no docs really on what's required and how it works. Just running `cbindgen -l c` as comment says outputs for me quite a different file. I could use some help 🙏 

Fix #325

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 
 t/servroot
 Cargo.lock
+
+.idea/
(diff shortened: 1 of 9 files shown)


jakubdyszkiewicz:
closing as I don't have time to work on this and we switched to CEL. We can reopen if needed.

PR #336, opened 2026-01-19
Make c bindings actually generated

Dr-Emann:
Add a script to regenerate the bindings in `cdefs.lua` with cbindgen, like the comment above hints at.

Related to #328, which talked about the fact that the `cdefs.lua` bindings say they're generated, but had no way to regenerate them.


src/ffi/expression.rs
```diff
@@ -49,7 +49,7 @@ impl Expression {
 
 bitflags! {
     #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-    #[repr(C)]
+    #[repr(transparent)]
```
Oyami-Srk: Any reason not to use `repr(C)`?

Dr-Emann:
Think this could also be nice for #328, to generate the bindings rather than presumably having to manually update the bindings.

Dr-Emann:
Ha, meant to mention #344, not 328 that I already mentioned.

cbindgen.toml
```diff
@@ -1,19 +1,26 @@
...
+no_includes = true
 
 [enum]
-prefix_with_name = true
+prefix_with_name = false
 
 [defines]
-"feature = ffi" = "DEFINE_ATC_ROUTER_FFI"
+# bindings only make sense if ffi is enabled
+# "feature = ffi" = "DEFINE_ATC_ROUTER_FFI"
```
Oyami-Srk, line 11: Could you please explain these? Thanks in advance!

Oyami-Srk (approved):
LGTM. Left a minor question.

cbindgen.toml
Dr-Emann: The prefix_with_name = false changes how enum values are named in the generated C code.

e.g. we generate

```c
typedef enum Type {
  String,
  IpCidr,
  IpAddr,
  Int,
  Regex,
} Type;
```

if prefix_with_name were still true, this would have been generated as:

```c
typedef enum Type {
  Type_String,
  Type_IpCidr,
  Type_IpAddr,
  Type_Int,
  Type_Regex,
} Type;
```

instead.

Dr-Emann: Removing the define for the ffi feature cfg means every single definition won't be wrapped in defines (which luajit doesn't understand), e.g.:

```c
#if defined(DEFINE_ATC_ROUTER_FFI)
struct Context *context_new(const struct Schema *schema);
#endif

#if defined(DEFINE_ATC_ROUTER_FFI)
void context_free(struct Context *context);
#endif
```

Because the _existence_ of the ffi API is based on the ffi feature being on, it doesn't make sense to require a define for it in the c bindings: it's intended for "function foo is only available if the 'foo' function is enabled", or "this cfg changes these functions from returning u32 to u64".

PR #340, opened 2026-02-02
style: remove legacy extern crate declarations

minxuanz:
After rust edition 2018, `use` statements can import procedural macro
https://doc.rust-lang.org/book/ch20-05-macros.html


Oyami-Srk (approved):
LGTM, Thanks for your contribution! 🚀

PR #341, opened 2026-02-03
Allow prefiltering based on string fields

### LINKED ISSUE #338: match_mix benchmark seems misleading
The code for the match_mix benchmark is at https://github.com/Kong/atc-router/blob/dc322f691f9d1e819315b2de5ba0899710734ff5/benches/match_mix.rs#L29-L30

I found the grouping in the format! call to be misleading: It looks like it should be parsed as:
```
(maybe true based on path) || (always true (a is never equal to both 2 and 9)) || ... || ...
```
which would logically simplify to
```
(always true)
```

However, in expressions, `||` binds more tightly than `&&`, and the last format argument has an `&&`, the expression therefore parses as:

```
((maybe true based on path) || (always true (a is never equal to both 2 and 9)) || ... || ...) && (!(a == 5))`
```

which logically simplifies to
```
(always true) && a != 5
```

Further, in the Matching benchmark:
https://github.com/Kong/atc-router/blob/dc322f691f9d1e819315b2de5ba0899710734ff5/benches/match_mix.rs#L42-L44

The use of `hello49999` as the http.path would seem to indicate that it's trying to match the middle expression in the list, but it's actually measuring the time to match against the very first matcher, just based on `a` not being equal to 5

### LINKED ISSUE #339: Prefilter based on http.path
I've been thinking of a way to speed things up for huge numbers of routes, by finding required prefixes for `http.path` for each expression, and using them to quickly narrow down the possible set of expressions to try, hopefully by a lot.

e.g. the following expressions:

```
[1]: net.dst.port == 9000
[2]: http.path ^= "/abc"
[3]: http.path ~ "^/foobar/(.*)$" && net.dst.port < 8000
[4]: http.path ~ "^/abc123" || http.path == "/123/456"
[5]: http.path ~ "^/123" && net.dst.port == 9001
```

We could preprocess that to have a mapping of possible prefixes to expressions to check:

(note that expression 1 can match regardless of the path, so it must always be considered)
```
"/abc" => [1, 2]
"/abc123" => [1, 2, 4]
"/foobar/" => [1, 3]
"/123" => [1, 5]
"/123/456" => [1, 4, 5]
otherwise => [1]
```
(where the longest prefix that is a prefix of the incoming context determines what expressions need to be checked)

This quick prefilter check should be able to reduce the number of required expression checks by quite a lot, with larger numbers of expressions, assuming _most_ expressions are at least partly based on the `http.path`.

Rust's `regex-syntax` crate [has the ability to extract required literal prefixes](https://docs.rs/regex-syntax/0.8.8/regex_syntax/hir/literal/index.html), so we can do that check even for regexes.

I've started working on [a library](https://github.com/Dr-Emann/router-prefilter) to do exactly this, and results so far seem promising.

Is this something that would be interesting to integrate here?

Dr-Emann:
Closes #339
Fixes #338

**Note, this requires calling the `enable_prefilter` method, otherwise nothing changes.**

In the case where most matches can be filtered out based on prefixes, it can greatly speed up matching by quickly eliminating large swaths of possible expressions.

I believe most of the time, the `http.path` field will be ideal for this filtering, where most of the time, looking at the prefix of the value can narrow down matching to just a small number of possible matches.

It makes building the router a bit slower

```
Build Router            time:   [4.3063 ms 4.3118 ms 4.3177 ms]
Build Router with Prefilter
                        time:   [5.0742 ms 5.0842 ms 5.0949 ms] (1.18x)
```

But it can _greatly_ decrease the cost of matching on all but the very first few expressions:

```
match_mix/match/0       time:   [248.85 ns 249.38 ns 250.00 ns]
match_mix/match/10      time:   [525.80 ns 526.30 ns 526.99 ns]
match_mix/match/49999   time:   [15.977 ms 16.450 ms 16.946 ms]
match_mix/match/99999   time:   [32.377 ms 33.620 ms 34.941 ms]
match_mix/no match/100001
                        time:   [30.529 ms 31.400 ms 32.325 ms]

match_mix with prefilter/match/0
                        time:   [337.58 ns 338.34 ns 339.10 ns] (1.36x)
match_mix with prefilter/match/10
                        time:   [408.87 ns 410.04 ns 411.50 ns] (0.779x)
match_mix with prefilter/match/49999
                        time:   [578.56 ns 579.68 ns 580.93 ns] (0.0000352x)
match_mix with prefilter/match/99999
                        time:   [616.64 ns 617.38 ns 618.24 ns] (0.0000184x)
match_mix with prefilter/no match/100001
                        time:   [379.15 ns 379.83 ns 380.60 ns] (0.0000121x)
```


Oyami-Srk:
That was impressive! Did you mean that the old method for `match_mix/match/99999` runs about 33 ms, and with prefiltering, it goes down to 617.38 ns?

Dr-Emann:
Yes, effectively. The benchmarking is in this PR: It sets up 100_000 routes that look like:
`(http.path == "hello{i}" && http.version == "1.1") && ({always_true} || {...} || {...}) && !(a == 5)`. It then makes a request for `hello99999` against a router that is and is not configured with the prefilter on `http.path`. Without a prefilter it has to try every route from 0 until it finds one that matches. With the prefilter, it only has to try a few routes.

In that case, it ends up having to try the routes that match against `hello9`, `hello99`, `hello999`, `hello9999`, and finally `hello99999` (in that order), and finds a match after only actually testing against 5 routes rather than testing 100000 routes, which is why it's so many times faster.

Dr-Emann:
I added some benchmarks for the worst case, and fleshed out some of the existing benchmarks.

### More Realistic Benchmarks

#### Building with/without prefilter
build-router-comparison

#### Matching with/without prefilter
match_mix_comparison


### Worst Case Benchmarks

I wanted to compare in the worst case for the prefilter, where the prefixes found are _all_ prefixes of each other (starts with `/a` or starts with `/a/a` or starts with `/a/a/a` or ...)

#### Building with/without prefilter
worst-case-build

#### Matching with/without prefilter
worst-case-match

Oyami-Srk:
Thanks for your huge contribution! I'm wondering if we can fork [router-prefilter](https://github.com/Dr-Emann/router-prefilter) to our organization or make it a sub-crate to this repo to align with several compliance requirements.
cc @michaelxiong-byte, also @StarlightIbuki for performance benchmark with Kong.

Dr-Emann:
Sure, and totally understandable. I'd be open to _either_ a fork or incorporating as a sub-crate directly in this repo, whichever works best for you.

Oyami-Srk:
Personally, I think sub-crate works the best for this situation, if it's not too much trouble. I'm reviewing the router_prefilter library, and the overall structure looks good to me. I haven't fully gone through all the pieces yet, and just FYI, I'll be out for the next 2 weeks, so perhaps low responsive during that time.

StarlightIbuki:
This should be effective. However, I need to find a more representative test suite to verify. Working on it

StarlightIbuki:
Test results show high confidence in the improvement. I used GitHub OpenAPI spec and random requests as the test scenario. It shows a slight, but statistically significant improvement in throughput for a constant config(which is kind of expected as cache plays a role in this case); and it shows great improvement in P99 (tail latency) for config changes(route configuring during the tests).
The plot is on the way.

Dr-Emann:
> I used GitHub OpenAPI spec and random requests as the test scenario
> ... 
> which is kind of expected as cache plays a role in this case

What does "random requests" mean in this case? Totally random urls, or urls with a random component (/users/{uuid}) (cache wouldn't work in that case)? Or maybe random choices from a set of matching urls?

StarlightIbuki:
> > I used GitHub OpenAPI spec and random requests as the test scenario
> > ...
> > which is kind of expected as cache plays a role in this case
> 
> What does "random requests" mean in this case? Totally random urls, or urls with a random component (/users/{uuid}) (cache wouldn't work in that case)? Or maybe random choices from a set of matching urls?

URL towards random valid APIs, with random valid parameters.

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,457 @@
+use bstr::BString;
+use std::collections::{BTreeMap, BTreeSet};
+use std::iter;
+use std::mem;
+
+#[derive(Debug, Clone)]
+struct RadixTrie<K> {
```
Oyami-Srk, line 7: Have you considered using some existing radix trie library, such as https://crates.io/crates/radix_trie?

StarlightIbuki:
Config updating (changing router)
changeroute conf-github script-github default_major

Constant config
defualt conf-github script-github default_major

crates/atc_router_prefilter/src/matchers.rs
```diff
@@ -0,0 +1,660 @@
...
+        let did_insert = lhs_inner.insert(v);
+        len += usize::from(did_insert);
+        if len > max_len {
+            *lhs = None;
+            return;
+        }
+    }
+}
+
+fn intersect_prefix_expansions_into(
```
Oyami-Srk, line 449: IIUC, `intersect_prefix_expansions_into` calculates the longest common prefix between `lhs` and `rhs`. Could you provide some comments for these helper functions in order to help reviewers?

For the simplest case, when `lhs = { a, base }` and `rhs = { api, b }`, the result should be `{ api, base }`.

However, due to the early return mechanism, it seems like the algorithm is somehow "greedy", for example, if `lhs = { a, apple }` and `rhs = { app }`, the result will be `{ app }` instead of `{ apple }`.

Could you please clarify this?

StarlightIbuki:
Lax Max is highly dominated by randomness. Lat P99 and P90 are better indicators for tail latencies.

src/router.rs
```diff
@@ -1,21 +1,31 @@
...
 use uuid::Uuid;
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 struct MatcherKey(usize, Uuid);
 
+#[derive(Debug)]
+struct PrefilteredField {
+    field: String,
+    prefilter: RouterPrefilter<Reverse<MatcherKey>>,
```
Oyami-Srk, line 23: Remark: the reason for using `Reverse` here is that originally router iterates over `self.matchers.iter().rev()`, so it's reversed.

The `BTreeMap` and `BTreeSet` should share the same ordering, so the execution order will remain the same; therefore, there should be no breaking change.

crates/atc_router_prefilter/src/matchers.rs
Oyami-Srk: Actually, I think the problem is that I'm not super clear about the common prefix in the same group when doing the intersection.

Dr-Emann: Added some comments that will hopefully help explain.

Dr-Emann:
I wonder how CPU usage is affected, as well. The Kong blog [has said](https://konghq.com/blog/engineering/mysterious-performance-regression):

> Router matching represents one of the most resource-intensive operations within our core proxy path

src/router.rs
Dr-Emann: Added a comment to this effect as well

crates/atc_router_prefilter/src/matchers.rs
Oyami-Srk: Thanks! Now it's clearer. So the greedy calculation of the prefix is intended.

crates/atc_router_prefilter/src/inner_prefilter.rs
Dr-Emann: Some, yeah.

That one, for example:
1. [Stores](https://github.com/michaelsproul/rust_radix_trie/blob/a89f789e126741a1814109b6ac15fff7c143a077/src/trie_node.rs#L15) the radix trie keys (path) explicitly in leaves, rather than like we do, where the key is implicitly stored in the path to the trie node. (this is effectively required because it's generic over the key type).
2. We need to collect all nodes which store a prefix of the searched path. That library doesn't give the ability to navigate the radix trie directly, so I think the best option we would have would be to walk backwards (repeatedly re-searching), something like:
```rust
while let Some(subtrie) = self.trie.get_ancestor(path_search) {
    // If get_ancestor returns Some, the subtree must have a key/value
    let key = subtrie.key().unwrap();
    let set = subtrie.value().unwrap();
    result.extend(set);
    let Some(len) = key.len().checked_sub(1) else {
        break;
    };
    path_search = &path_search[..len];
}
```

Oyami-Srk: Make sense. Thanks!

StarlightIbuki:
> I wonder how CPU usage is affected, as well. The Kong blog [has said](https://konghq.com/blog/engineering/mysterious-performance-regression):
> 
> > Router matching represents one of the most resource-intensive operations within our core proxy path

The recorded CPU usage shows 100% utilization. The test is CPU-bound, thus it's already reflected in the throughput metrics.
It requires a different test scenario to check how the CPU usage changes

src/router.rs
```diff
@@ -87,22 +108,172 @@ where
...
 
         None
     }
+
+    /// Enable prefiltering on the specified field.
+    pub fn enable_prefilter(&mut self, field: &str) {
+        match self.schema.borrow().type_of(field) {
+            Some(Type::String) => {}
+            Some(actual) => panic!("Field {field} is of type {actual:?}, must be a string"),
+            None => panic!("Field {field} is not in schema"),
```
Oyami-Srk: Shall we use error here?

src/router.rs
```diff
@@ -87,22 +108,172 @@ where
...
+        }
+        let mut prefilter = RouterPrefilter::new();
+        for (key, expr) in &self.matchers {
+            prefilter.insert(Reverse(key.clone()), ExprMatcher { expr, field });
+        }
+        self.prefiltered_field = Some(PrefilteredField {
```
Oyami-Srk, line 196: What if we call this method multiple times? I think this is worth a guard or documentation to avoid misuse.
BTW, I think it's also worth mentioning that this method is not like flipping a boolean flag; it will build a prefilter, so it will be slow. (and also applies for readme)

Oyami-Srk (commented):
Overall LGTM. Could you please extend the existing tests to use `enable_prefilter`?

Oyami-Srk (approved):
LGTM, still need some review and decision from the Kong 3 team. cc @michaelxiong-byte

Dr-Emann:
I've made some changes to `router_prefilter` upstream that might be useful for either debug logging, or possibly for https://github.com/Kong/atc-router-wasm, separating out prefix extraction from matchers and inserting into a prefilter. I don't know if we want to bring them in.

Dr-Emann:
Anything I can help with to move this along any?

VicYP:
> Anything I can help with to move this along any?

crates/atc_router_prefilter/scripts/update_github_urls_for_bench.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+url="https://raw.githubusercontent.com/github/rest-api-description/04fd6c592fc546217404b07e0b0e581fb00a963a/descriptions/api.github.com/api.github.com.json"
+
+curl -sSL "$url" | jq -r '.paths | to_entries[] | .key as $path | .value | keys[] | "\(.):\($path)"' > benches/github_paths.txt
```
StarlightIbuki: I don't see a user of this script, and the bench requires "github_paths.txt" to exist. Should we automate this? Or under what circumstances do we need to run this manually?

StarlightIbuki: Also, a nitpick: 
```suggestion
curl -sSL "$url" | jq -r '.paths | to_entries[] | .key as $path | .value | keys[] | "\(.):\($path)"' > benches/github_paths.txt

```

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,457 @@
...
+    keys: BTreeSet<K>,
+    children: Vec<RadixLink<K>>,
+}
+
+#[derive(Debug, Clone)]
+struct RadixLink<K> {
```
StarlightIbuki: The data structure design introduces heavy indirection with pointers.
Maybe we can avoid this with a different data structure to get better memory locality and a higher cache hit rate?
Say we shift to use index-based allocation of nodes instead of placing them around the heap; or maybe let's utilize an arena crate.

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,457 @@
+use bstr::BString;
+use std::collections::{BTreeMap, BTreeSet};
+use std::iter;
+use std::mem;
+
+#[derive(Debug, Clone)]
+struct RadixTrie<K> {
+    keys: BTreeSet<K>,
+    children: Vec<RadixLink<K>>,
```
StarlightIbuki: With a layout like this, it helps with memory locality:

```suggestion
    // associate with index
    children_initials: Vec<u8>,
    children_rests: Vec<BString>,
    children: Vec<RadixTrie<K>>,
```

The code block is merely demonstrating.

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,457 @@
...
+}
+
+#[derive(Debug, Clone)]
+struct RadixLink<K> {
+    ch: u8,
+    rest: BString,
```
StarlightIbuki: Maybe we can try "SmallVec"?

crates/atc_router_prefilter/src/inner_prefilter.rs
StarlightIbuki: And if we want to shift to the flat data structure, we may replace the children Vec's item with the outer vector's index

StarlightIbuki: And I suppose this will speed up scanning significantly; it will also unlock further optimization when searching for the node (say, a better lookup algorithm for initials)

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,457 @@
...
+        // Clean up empty nodes.
+        if link.child.keys.is_empty() && link.child.children.is_empty() {
+            self.children.remove(idx);
+        } else {
+            try_compact_link(&mut self.children[idx]);
+        }
+    }
+
+    fn collect_prefix_matches(&self, mut input: &[u8]) -> BTreeSet<&K> {
+        let mut result = BTreeSet::new();
```
StarlightIbuki, line 109: We may improve this: This allocates memory every time we look up; we may memoize this to avoid recalculation(and allocation).
And it does not need to be a BTreeSet; we can just collect all the keys and sort them once. A possibly more aggressive optimization is to use a bucket, as it deduplicates naturally. We will have a limited number of UUIDs and priorities; thus it's feasible. A bit set also works well here if we find a way to map it to dense integers

StarlightIbuki: Given that there will be more reads compared to writes, I find it worthwhile to shift the cost to the updating of the structure

Cargo.toml
```diff
@@ -42,3 +44,10 @@ harness = false
 [[bench]]
 name = "match_mix"
 harness = false
+
+[[bench]]
+name = "worst_case"
+harness = false
+
+[workspace]
+members = ["crates/*"]
```
StarlightIbuki: Nitpick:
```suggestion
members = ["crates/*"]

```

crates/atc_router_prefilter/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "atc_router_prefilter"
+version = "1.3.3"
+edition = "2024"
```
StarlightIbuki, line 4: This conflicts with the FFI wrapper, which uses 2021; also, this needs extra care when building Kong with this PR. Also, I did not check the reason, but the dependencies require massive bumping.

src/router.rs
```diff
@@ -87,22 +108,189 @@ where
...
+        context: &'a Context,
+    ) -> Option<RouterPrefilterIter<'a, Reverse<MatcherKey>>> {
+        match &self.prefiltered_field {
+            Some(PrefilteredField { field, prefilter }) if prefilter.can_prefilter() => {
+                let values = context.value_of(field)?;
+                let value = values.first()?;
+                let value = value.as_str()?;
+                Some(prefilter.possible_matches(value))
+            }
+            _ => None,
```
StarlightIbuki: Just my personal preference:

```suggestion
        if prefilter.can_prefilter() return None;
        let Some(PrefilteredField  { field, prefilter }) = &self.prefiltered_field  else {
            return None;
        };

        let values = context.value_of(field)?;
        let value = values.first()?;
        let value = value.as_str()?;
        Some(prefilter.possible_matches(value))
```

This way the logic is flattened.

src/router.rs
```diff
@@ -87,22 +108,189 @@ where
...
+        self.prefilter.remove(&Reverse(key));
+    }
+}
+
+struct ExprMatcher<'a> {
+    expr: &'a Expression,
+    field: &'a str,
+}
+
+impl Matcher for ExprMatcher<'_> {
```
StarlightIbuki, line 230: Just noting to myself: maybe we need to refactor to avoid this boilerplate

src/router.rs
```diff
@@ -87,22 +108,189 @@ where
...
+    fn visit(&self, visitor: &mut MatcherVisitor) {
+        match self.expr {
+            Expression::Logical(logical) => {
+                visitor.visit_nested_start();
+                match logical.as_ref() {
+                    LogicalExpression::And(lhs, rhs) => {
```
StarlightIbuki, line 236: Nitpick and personal preference again: we may introduce a helper function to extract the "left_matcher" and "right_matcher"

crates/atc_router_prefilter/benches/router_bench.rs
```diff
@@ -0,0 +1,154 @@
...
+            let (method, path) = line
+                .split_once(':')
+                .expect("every line should have a colon");
+            (method.to_string(), path.to_string())
+        })
+        .collect()
+}
+
+fn path_to_regex(path: &str) -> String {
+    static RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"\{[^}]*}").unwrap());
```
StarlightIbuki, line 29: Note to myself: maybe at some point we need a comment to let the linter ignore "unwrap()"

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,457 @@
...
+
+fn try_compact_link<K>(link: &mut RadixLink<K>) {
+    if link.child.keys.is_empty() && link.child.children.len() == 1 {
+        let grandchild = link.child.children.pop().unwrap();
+        link.rest.reserve(1 + grandchild.rest.len());
+        link.rest.push(grandchild.ch);
+        link.rest.extend_from_slice(&grandchild.rest);
+        link.child = grandchild.child;
+    }
+}
```
StarlightIbuki: Again, my personal preference

```suggestion
fn try_compact_link<K>(link: &mut RadixLink<K>) {
    if !link.child.keys.is_empty() || link.child.children.len() != 1 {
        return;
    }
    // make the linter happy
    let Some(grandchild) = link.child.children.pop() else {
        return;
    };

    link.rest.reserve(1 + grandchild.rest.len());
    link.rest.push(grandchild.ch);
    link.rest.extend_from_slice(&grandchild.rest);
    link.child = grandchild.child;
}
```

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,457 @@
...
+
+fn split_link<K>(link: &mut RadixLink<K>, at: usize) {
+    let tail = link.rest.split_off(at + 1);
+    let ch = link.rest.pop().unwrap();
```
StarlightIbuki: The same technique applies here.
Also, do we have a guarantee of the condition? It seems that we are actually postponing the assertion, which is a conditional branch there:

```lua
            if common_len < link.rest.len() {
                split_link(link, common_len);
            }
```

StarlightIbuki: IMO, this should be a precondition instead of an in-the-middle assertion

src/ffi/router.rs
```diff
@@ -147,6 +147,73 @@ pub unsafe extern "C" fn router_remove_matcher(
...
+#[no_mangle]
+pub unsafe extern "C" fn router_enable_prefilter(
+    router: &mut Router<&Schema>,
+    field: *const u8,
+    errbuf: *mut u8,
+    errbuf_len: &mut usize,
+) -> bool {
+    let field = ffi::CStr::from_ptr(field as *const c_char)
+        .to_str()
+        .unwrap();
```
StarlightIbuki, line 194: This `unwrap()` needs to be justified. It should also be handled.

src/router.rs
```diff
@@ -87,22 +108,189 @@ where
...
     }
 
+    fn prefilter_matches<'a>(
+        &'a self,
+        context: &'a Context,
+    ) -> Option<RouterPrefilterIter<'a, Reverse<MatcherKey>>> {
+        match &self.prefiltered_field {
+            Some(PrefilteredField { field, prefilter }) if prefilter.can_prefilter() => {
+                let values = context.value_of(field)?;
+                let value = values.first()?;
```
StarlightIbuki: We need a comment for this. It's not trivial to find that this does not alter the behavior

StarlightIbuki (commented):
Generally a solid PR.

A few notable things:
1. An `unwrap()` call at the FFI boundary needs to be justified;
2. It bumps the edition, making it inconsistant to the FFI wrapper;
3. The size of the PR itself is huge and introduces multiple new dependencies. We need to ensure it's opt-in, and yet we need to consider either maintaining a new branch or merging it to the main branch;4. I left some comments for possible improvement of the data structure. Basically, we need to take care of memory locality
Others are personal preference and not blockers.

src/router.rs
```diff
@@ -87,22 +108,189 @@ where
...
+                visitor.visit_nested_finish();
+            }
+            Expression::Predicate(pred) => {
+                if pred.lhs.var_name == self.field && pred.lhs.transformations.is_empty() {
+                    match pred.op {
+                        BinaryOperator::Equals => {
+                            let rhs = pred
+                                .rhs
+                                .as_str()
+                                .expect("can only use a prefilter on strings");
```
StarlightIbuki, line 274: Those `expect()` may break through the FFI boundary and causes crash

src/router.rs
StarlightIbuki: This relies on `transformations.is_empty()`: https://github.com/dr-emann/atc-router/blob/22b3bb2ce95ae0e1ebf7d8743c752a1136531e33/src/router.rs#L257

src/router.rs
```diff
@@ -87,22 +108,189 @@ where
...
         None
     }
+
+    /// Enable prefiltering on the specified field.
+    ///
+    /// This will compile a prefilter for all currently existing matchers, and all future added
+    /// matchers will be added to the prefilter. This can be an expensive call if there are a lot
+    /// of matchers.
+    ///
+    /// Calling
```
StarlightIbuki: Unfinished comment

crates/atc_router_prefilter/src/lib.rs
```diff
@@ -0,0 +1,840 @@
...
+/// for (i, route) in routes.into_iter().enumerate() {
+///     prefilter.insert(i, route);
+/// }
+/// let matches: Vec<_> = prefilter.possible_matches("/api/posts").collect();
+/// assert!(matches.contains(&&0));
+/// ```
+#[derive(Debug)]
+pub struct RouterPrefilter<K> {
+    // Only includes indexes after prefilter starts
+    always_possible: BTreeSet<K>,
```
StarlightIbuki, line 54: Note to myself: needs further inspection

crates/atc_router_prefilter/Cargo.toml
Dr-Emann: I'm not sure what you mean by it conflicting with the FFI wrapper. Do you need to support a version of rust that doesn't support the 2024 edition? You can use crates with different editions together.

I imagine it wouldn't be hard to use the 2021 edition instead, but I'd like to know your minimum rust version to support, to be sure I'm not using any APIs that wouldn't exist yet on your compiler, if that's the case.

src/router.rs
Dr-Emann: Can't be rewritten exactly that way since `prefilter` doesn't exist until after the destructuring, but can rewrite as:

```rust
let PrefilteredField { field, prefilter } = self.prefiltered_field.as_ref()?;
if !prefilter.can_prefilter() {
    return None;
}
let values = context.value_of(field)?;
let value = values.first()?;
let value = value.as_str()?;
Some(prefilter.possible_matches(value))
```

src/router.rs
Dr-Emann: These depend on the correctness of the expression parser, we only do prefiltering on types where the schema says it's a string:

https://github.com/Dr-Emann/atc-router/blob/446287a392405d707b2bdb3a6912a6e2dbc804cf/src/router.rs#L172-L180

And rely on values with that field name to be required to be strings

https://github.com/Dr-Emann/atc-router/blob/446287a392405d707b2bdb3a6912a6e2dbc804cf/src/router.rs#L257

Dr-Emann:
> introduces multiple new dependencies

This should actually introduce exactly one new runtime dependency: `bstr`. I could remove that, if we'd like, it's just used for nicer debug printing of byte strings. It's written by burntsushi, the author of the regex crates, and is pretty self-contained, though.

Before:

```
atc-router v1.7.1 (/Users/zach/Development/tmp/atc-router)
├── bitflags v2.10.0
├── cidr v0.3.2
├── fnv v1.0.7
├── lazy_static v1.5.0
├── pest v2.8.6
│   ├── memchr v2.8.0
│   └── ucd-trie v0.1.7
├── regex v1.12.3
│   ├── aho-corasick v1.1.4
│   │   └── memchr v2.8.0
│   ├── memchr v2.8.0
│   ├── regex-automata v0.4.14
│   │   ├── aho-corasick v1.1.4 (*)
│   │   ├── memchr v2.8.0
│   │   └── regex-syntax v0.8.9
│   └── regex-syntax v0.8.9
└── uuid v1.20.0
```

After:

```
atc-router v1.7.1 (/Users/zach/Development/tmp/atc-router)
├── atc_router_prefilter v1.3.3 (/Users/zach/Development/tmp/atc-router/crates/atc_router_prefilter)
│   ├── bstr v1.12.1
│   │   ├── memchr v2.8.0
│   │   └── regex-automata v0.4.14
│   │       ├── aho-corasick v1.1.4
│   │       │   └── memchr v2.8.0
│   │       ├── memchr v2.8.0
│   │       └── regex-syntax v0.8.9
│   └── regex-syntax v0.8.9
├── bitflags v2.10.0
├── cidr v0.3.2
├── fnv v1.0.7
├── lazy_static v1.5.0
├── pest v2.8.6
│   ├── memchr v2.8.0
│   └── ucd-trie v0.1.7
├── regex v1.12.3
│   ├── aho-corasick v1.1.4 (*)
│   ├── memchr v2.8.0
│   ├── regex-automata v0.4.14 (*)
│   └── regex-syntax v0.8.9
└── uuid v1.20.0
```

crates/atc_router_prefilter/scripts/update_github_urls_for_bench.sh
Dr-Emann: It can probably just be removed. When possible, I like to include the script used to generate data, but it seems unlikely that it'd be useful to ever run it again.

crates/atc_router_prefilter/src/inner_prefilter.rs
Dr-Emann: The problem with those was the need to update the prefilter incrementally.

If kong would be able to use an API where `enable_prefilter` needs to be re-called after every batch of updates (prefilter would be automatically disabled after any modification to the router), we could definitely make some different choices.

src/ffi/router.rs
Dr-Emann: Following existing code's use of unwrap/expect at the ffi boundary for unicode strings. Agreed, should probably change all these, unsure if it should be part of this PR.

e.g. small sample of existing ffi entrypoints

- https://github.com/Kong/atc-router/blob/25a34f0c8a5824459160742785fcc8cfc14784a8/src/ffi/expression.rs#L163
- https://github.com/Kong/atc-router/blob/25a34f0c8a5824459160742785fcc8cfc14784a8/src/ffi/context.rs#L91-L93
- https://github.com/Kong/atc-router/blob/25a34f0c8a5824459160742785fcc8cfc14784a8/src/ffi/router.rs#L101-L104

crates/atc_router_prefilter/src/inner_prefilter.rs
Dr-Emann: Funnily enough, thought about it, but wasn't sure about adding a new dependency.

Would be happy to swap to using smallvec if we'd like.

src/router.rs
Dr-Emann: Added a comment that hopefully clarifies

Dr-Emann:
About using a separate branch vs merging to main, I'm not _against_ a separate branch, but I'm not sure it's meaningfully worth it.

The vast, vast majority of the changes have no effect if `enable_prefilter` is never called. I'd recommend reviewing [router.rs](https://github.com/Kong/atc-router/pull/341/changes#diff-08d36d8c77bb769044ed645092d1240f6e4c15187866104347da55c3709017b7) carefully for what happens in case the prefilter is never called:

The `prefiltered_field` is only set to Some inside the `enable_prefilter` method.
The changes to `add_matcher_expr` and `remove_matcher` are both no-ops if `prefiltered_field` is Some.
`try_match` calls `prefilter_matches`, which always returns None if `prefiltered_field` is None, and `try_match` acts identically to before if `prefilter_matches` returned None.

So while it's quite a big PR, I think it's so nearly all opt-in that I don't see a huge benefit to keeping it as a longer lived feature branch or anything.

StarlightIbuki:
One more noticeable thing: we are enabling the prefilter for most of the tests. Should we consider making it a test matrix to cover both paths?

crates/atc_router_prefilter/src/inner_prefilter.rs
StarlightIbuki: Definitely worth considering. Or we may update by inserting (given it's a relatively fast operation, and we read more than we write), or even allowing holes in arena (won't be much)

crates/atc_router_prefilter/Cargo.toml
StarlightIbuki: We are currently using rustc-1.82.0, and I bumped a lot of our dependencies to keep up with it. Confirming with the team about this.

crates/atc_router_prefilter/src/inner_prefilter.rs
StarlightIbuki: Let's perf it then. If it works, let's swap

crates/atc_router_prefilter/Cargo.toml
StarlightIbuki: Generally, I'm OK with bumping MSRV but we need to be sure it won't be a surprise

crates/atc_router_prefilter/src/inner_prefilter.rs
StarlightIbuki: and the vec is not neccessarily sorted, or we may use a pooll; you name it. a lot of techniques can be applied

crates/atc_router_prefilter/src/inner_prefilter.rs
StarlightIbuki: Still I prefer column-based table(struct of vec) over row-based table(OO classes, vec of struct)

StarlightIbuki:
Sorry for the misleading perf results. Before reviewing, I did not realize the image I created did not actually enable the perfilter

src/ffi/router.rs
StarlightIbuki: Sure. Maybe I will draft a PR for this

Dr-Emann:
> Sorry for the misleading perf results. Before reviewing, I did not realize the image I created did not actually enable the perfilter

I _was_ pretty suspicious that with truly random requests that there wasn't a bigger impact, since that should make the cache a lot less useful.

Dr-Emann:
> One more noticeable thing: we are enabling the prefilter for most of the tests. Should we consider making it a test matrix to cover both paths?

Not super familiar with the `prove` testing framework here. I don't see a clean way to matrix this without duplicating a lot of blocks. Is there a standard way to do this test matrix-ing?

StarlightIbuki:
While validating this branch against Kong/kong-ee's Bazel build (pinned to stable rustc 1.85.0, tracked in kong-ee PR Kong/kong-ee#20156), the build failed because \`inner_prefilter.rs\` uses \`slice::split_off_first()\`, which is a nightly-only unstable API and doesn't exist on stable Rust:

\`\`\`
error[E0599]: no method named `split_off_first` found for reference `&[u8]` in the current scope
  --> crates/atc_router_prefilter/src/inner_prefilter.rs:36:46
\`\`\`

Opened a fix here: Dr-Emann#1 — swaps the three call sites for the stable `split_first()` + explicit reassignment (behaviorally equivalent). All non-doctest tests pass; verified locally with `cargo build`/`cargo test` on stable rustc.

Dr-Emann:
Merged, and added a `rust-toolchain.toml` (and use older criterion version to work with rust 1.85) which will make e.g. CI run tests with rust 1.85, so usage of APIs which aren't yet stable in 1.85 can't sneak in.

crates/atc_router_prefilter/README.md
```diff
@@ -0,0 +1,50 @@
+# atc_router_prefilter
+
+Kong's fork of [`router_prefilter`].
```
StarlightIbuki, line 3: Should we create a fork? Or maybe let's just merge it to the original repo or make a 1.85.0 branch of it? It seems we do not have much actual diff from the original repo

StarlightIbuki (approved):
I'm okay with the feature. However I would still recommend maintaining a branch rather than merging: the gateway has cache and it makes the RPS benefits invisible in most cases. I will leave the decision to who merges PRs

crates/atc_router_prefilter/README.md
Dr-Emann: Forked based on https://github.com/Kong/atc-router/pull/341#issuecomment-3889567171

Dr-Emann:
Anything further I can do to help?

StarlightIbuki:
> Anything further I can do to help?

Hi. Given this is a really large PR, we need another pair of eyes to review, and it needs some dedicated hours. I'm raising this with people. Also, corrected perf tests show large improvement, but the test method itself is questioned. Sorry for being late and slow. Hopefully we are getting someone on it this way.

StarlightIbuki:
image
The perf result. We observed slightly increasement of memory usage and nearly 57x in RPS. (Note: this is a pure proxy test without any plugins)

crates/atc_router_prefilter/src/inner_prefilter.rs
```diff
@@ -0,0 +1,488 @@
...
+}
+
+impl<K: Ord> InnerPrefilter<K> {
+    /// Inserts a key with the given prefixes into the prefilter.
+    ///
+    /// Each prefix is added to the prefix map, maintaining the prefix-inheritance invariant.
+    ///
+    /// No prefix in `prefixes` may be a prefix of another entry in `prefixes`.
+    /// This precondition is upheld by the caller (`MatcherVisitor::finish`
+    /// applies `optimize_for_prefix_by_preference`, which collapses such
```
samugi: nit: Is this comment outdated? It looks like `MatcherVisitor::finish` is not calling this `optimize_for_prefix_by_preference` function.

src/router.rs
```diff
@@ -87,22 +108,195 @@ where
...
 
-        for (MatcherKey(_, id), m) in self.matchers.iter().rev() {
-            if m.execute(context, &mut mat) {
-                mat.uuid = *id;
-                return Some(mat);
+        match self.prefilter_matches(context) {
+            Some(possible_matches) => {
+                for key in possible_matches {
+                    let key = &key.0;
+                    let expr = &self.matchers[key];
```
samugi: looks like this one could panic if the key does not exist. Using `.get()` and handling `None` would be safer since we are in the hot path here.

README.md
```diff
@@ -165,6 +167,36 @@ matcher does not exist.
...
+
+**context:** *any*
+
+Enables prefiltering on the specified `field`. The field must be of type `String`
+in the router's schema. When enabled, the router uses the prefilter to narrow down
+candidate matchers before performing full evaluation, which can improve match
+performance.
+
+Calling this method requires building a prefilter for all current matchers: it may be 
+expensive. It is expected that this method will be called only once for a router. The
```
samugi, line 210: > It is expected that this method will be called only once for a router.

This reads as if calling it multiple times would be expensive, but it appears that is correctly handled and calling this multiple times on the same field is a noop

README.md
```diff
@@ -165,6 +167,36 @@ matcher does not exist.
...
+prefilter will be kept up to date with any added and removed matchers, until this method
+is called again with a different field name, or the disable_prefilter method is called.
```
samugi, line 212: I would also add explicitly that only one field can be prefiltered at a time.

README.md
```diff
@@ -165,6 +167,36 @@ matcher does not exist.
...
 [Back to TOC](#table-of-contents)
 
+### enable\_prefilter
+
+**syntax:** *res, err = r:enable_prefilter(field)*
+
+**context:** *any*
+
+Enables prefiltering on the specified `field`. The field must be of type `String`
+in the router's schema. When enabled, the router uses the prefilter to narrow down
```
samugi, line 177: also worth pointing out that not all expressions benefit from this optimization, it should be one of:

  * `==` (equals)
  * `^=` (prefix)
  * `~` with a pattern anchored at the start (`^` or `\A`, without the `m` flag)

And the field should be referenced directly (not through e.g. lower(...), etc).

It would be good to have a reference table that explains what are the requisites for a matcher to be prefilterable.

Also according to the benchmarks, it looks like this only pays off when there are many matchers, with distinct prefixes (and where most matchers are prefilterables as per the criteria above), I think it would be valuable to point this out in the docs (perhaps with some rough guidance on what's the amount of matchers we consider "enough" to make this setting worth it).

samugi (commented):
looks good to me, just a few comments, mostly about docs

README.md
Dr-Emann: I'm personally okay with under-promising here, unless Kong has a reason to call `enable_prefilter` multiple times with the same name, and it would be beneficial to document the guarantee that it's cheap to call again with the same field name.

crates/atc_router_prefilter/src/inner_prefilter.rs
Dr-Emann: Yep, removed this whole section of the comment, it's no longer true, and references a function that doesn't exist anymore 👍

README.md
samugi: no strong opinion on this one in particular, we can leave it as is

samugi (approved):

src/router.rs
```diff
@@ -87,22 +108,200 @@ where
...
-                mat.uuid = *id;
-                return Some(mat);
+        match self.prefilter_matches(context) {
+            Some(possible_matches) => {
+                for key in possible_matches {
+                    let key = &key.0;
+                    let Some(expr) = self.matchers.get(key) else {
+                        if cfg!(debug_assertions) {
+                            unreachable!("prefilter cannot return a matcher key not in `matchers`");
+                        }
```
Oyami-Srk, line 142: Worth mentioning that the debug guard may hide the real bug and silently cause a bug (by `continue` with a maybe-malformed prefilter).
Maybe we should be defensive and kill the prefilter when we meet this. Not a blocker for this PR, we can do it as a follow-up.

Oyami-Srk (approved):
Did an incremental review, LGTM still.
Maybe a question for Kong 3's usage: As we're using a lot of `debug_assert` in the code, do we correspondingly build unoptimized version of atc-router in Kong 3's CI to test the functionalities?

src/router.rs
Dr-Emann: Yeah. Not obvious how best do to that here, since we're in a function that takes `&self`.

bungle (approved):

PR #344, opened 2026-03-20
feat: add Bool type support across Rust and Lua APIs

chobits:
Use case: I'm adding field to context called `plugins.xxx.is_access_executed() == true/false` to indicate whether the some plugins in kong was executed. This is related to plugin conditional expression feature in kong.

### DIFF
--- a/lib/resty/router/cdefs.lua
+++ b/lib/resty/router/cdefs.lua
@@ -8,6 +8,7 @@ typedef enum Type {
   IpCidr,
   IpAddr,
   Int,
+  Bool,
   Regex,
 } Type;
 
@@ -22,6 +23,7 @@ typedef enum CValue_Tag {
   CValue_IpCidr,
   CValue_IpAddr,
   CValue_Int,
+  CValue_Bool,
 } CValue_Tag;
 
 typedef struct CValue_Str_Body {
@@ -41,6 +43,9 @@ typedef struct CValue {
     };
     struct {
       int64_t int_;
+    };
+    struct {
+      bool bool_;
     };
   };
 } CValue;
--- a/lib/resty/router/context.lua
+++ b/lib/resty/router/context.lua
@@ -63,6 +63,9 @@ function _M:add_value(field, value)
     elseif typ == "Bool" then
         CACHED_VALUE[0].tag = C.CValue_Bool
         CACHED_VALUE[0].bool_ = value
+
+    else
+        return nil, "cannot set context value for field of type " .. typ
     end
 
     local errbuf = get_string_buf(ERR_BUF_MAX_LEN)
--- a/lib/resty/router/context.lua
+++ b/lib/resty/router/context.lua
@@ -38,7 +38,7 @@ end
 
 
 function _M:add_value(field, value)
-    if not value then
+    if value == nil then
         return true
     end
 
@@ -59,6 +59,10 @@ function _M:add_value(field, value)
     elseif typ == "Int" then
         CACHED_VALUE[0].tag = C.CValue_Int
         CACHED_VALUE[0].int_ = value
+
+    elseif typ == "Bool" then
+        CACHED_VALUE[0].tag = C.CValue_Bool
+        CACHED_VALUE[0].bool_ = value
     end
 
     local errbuf = get_string_buf(ERR_BUF_MAX_LEN)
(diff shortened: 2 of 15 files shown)


src/ffi/expression.rs
```diff
@@ -218,7 +218,8 @@ pub unsafe extern "C" fn expression_validate(
 mod tests {
     use super::*;
     use crate::ast::Type;
-    use crate::ffi::ERR_BUF_MAX_LEN;
+
+    const ERR_BUF_MAX_LEN: usize = 4096;
```
chobits: this is not a good way to fix this issue, 
and also i foudn that our Ci workflow does not include "cargo test"

chobits: @Oyami-Srk

lib/resty/router/context.lua
```diff
@@ -59,6 +59,10 @@ function _M:add_value(field, value)
     elseif typ == "Int" then
         CACHED_VALUE[0].tag = C.CValue_Int
         CACHED_VALUE[0].int_ = value
+
+    elseif typ == "Bool" then
+        CACHED_VALUE[0].tag = C.CValue_Bool
+        CACHED_VALUE[0].bool_ = value
```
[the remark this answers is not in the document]
chobits: fixed in https://github.com/Kong/atc-router/pull/344/changes/5dc11230684a777787bf11815b3d2c78ae0cbdd2

src/parser.rs
```diff
@@ -95,7 +95,7 @@ fn parse_lhs(pair: Pair<Rule>) -> ParseResult<Lhs> {
     })
 }
 
-// rhs = { str_literal | ip_literal | int_literal }
+// rhs = { str_literal | ip_literal | int_literal | bool_literal }
```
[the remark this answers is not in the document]
chobits: fixed in https://github.com/Kong/atc-router/pull/344/commits/558fe2de7db9586fb1bf2dc46af4c02eabd62fa5

chobits:
based on this https://github.com/Kong/atc-router/pull/328, i re-implement boolean support

lib/resty/router/cdefs.lua
```diff
@@ -42,6 +44,9 @@ typedef struct CValue {
     struct {
       int64_t int_;
     };
+    struct {
+      bool bool_;
```
Oyami-Srk, line 48: According to [Rust's standard](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.validity.bool), the `bool` is either `0` or `1`; other values cause UB, and in the C99 standard, the value constraint for `_Bool` is applied during conversion:

> When any scalar value is converted to _Bool, the result is 0 if the value compares equal
> to 0; otherwise, the result is 1.

What I'm concerned about is whether values other than `0` and `1` are valid in `_Bool` in C99. If that is a UB, I think we should be fine.

Some related discussion:
- https://github.com/mozilla/cbindgen/issues/670
- https://github.com/rust-lang/rfcs/pull/954
- https://github.com/rust-lang/rfcs/issues/992
- https://en.cppreference.com/w/c/language/arithmetic_types.html#Boolean_type

Uh, this is just a corner case; no one actually cares too much. I'm fine with that.

Oyami-Srk (approved):
Implementation LGTM, introducing `boolean` type makes sense to me.

cc @michaelxiong-byte for Kong 3.
cc @javierguerragiraldez for Koko.
cc @michaelbeaumont @jakubdyszkiewicz for Event Gateway.
cc @sumimakito for Kong Manager (maybe we need to update Monaco editor to support new literal as well?).
cc @samugi for Houdini.
cc @dndx for language design of introducing a new type.

Huge thanks to everyone involved!

lib/resty/router/cdefs.lua
Oyami-Srk: C23 changed it a little bit, but no fundamental difference IMHO. I think it should be fine as long as we're not trying to exploit ourselves.

javierguerragiraldez:
IIUIC, this only adds the new Bool type and no new field; right?  IOW, the `feature.enabled` is only a test?
if so, this will need a similar PR on [go-atc-router](https://github.com/kong/go-atc-router)

After that, any new field of this type should just be added to the `expressions_compat.json` file, as usual.

chobits:
hi all, let's hold off on this PR for now. Considering that even a minor change here would require significant coordination and updates from upstream users, both Victor and the team prefer to keep things as they are for the moment.

We have a meeting with PM tomorrow to discuss whether we can avoid modifying the ATC router by using expression syntactic sugar instead. For example: `plugins.xxx.is_access_executed() == "true"` (using the string "true").

PR #347, opened 2026-07-16
feat(ffi): return matched expression in get_result

STARRY-S:
## Description

Add `matched_expressions` to `Match` struct and expose it through `context_get_result` FFI and lua `get_result`, so caller can get the original expression that matched a field.

Bump version to 1.8.0.

## Benchmark

Use `cargo bench --bench match_mix` to ensure no performance regression:

Screenshot 2026-07-21 at 11 55 03

## Fixes

[FTI-7643](https://konghq.atlassian.net/browse/FTI-7643)

[FTI-7643]: https://konghq.atlassian.net/browse/FTI-7643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "atc-router"
-version = "1.7.1"
+version = "1.8.0"
 edition = "2021"
```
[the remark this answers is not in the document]
STARRY-S: Kong pins atc-router library by commit, any upgrade is always an explicit.
This is not breaking, will not affect user experience.

lib/resty/router/context.lua
```diff
@@ -75,15 +75,17 @@ end
...
     if captures_len == -1 then
         return nil
     end
 
-    local matched_value_buf, matched_value_len
+    local matched_value_buf, matched_value_len, matched_expr_buf, matched_expr_len
     if matched_field then
         matched_value_buf = ffi_new("const uint8_t *[1]")
         matched_value_len = ffi_new("size_t [1]")
+        matched_expr_buf = ffi_new("const uint8_t *[1]")
```
StarlightIbuki, line 87: we may save the copy of string if we use a constant string pointer here. is that possible?

StarlightIbuki (approved):
generally good to me

src/interpreter.rs
```diff
@@ -49,6 +49,15 @@ impl Execute for Predicate {
...
                             .insert(self.lhs.var_name.clone(), self.rhs.clone());
+                        let rhs_str = match &self.rhs {
+                            Value::String(s) => s.clone(),
+                            Value::IpCidr(cidr) => cidr.to_string(),
+                            Value::IpAddr(addr) => addr.to_string(),
+                            Value::Int(i) => i.to_string(),
+                            Value::Regex(r) => r.as_str().to_string(),
+                        };
+                        m.matched_expressions
+                            .insert(self.lhs.var_name.clone(), rhs_str);
```
chobits: for equals, `matched_value` is already the rhs, so `matched_expression` just duplicates it here

lib/resty/router/context.lua
```diff
@@ -99,15 +101,19 @@ function _M:get_result(matched_field)
...
                             capture_values_len)
 
     local uuid = ffi_string(UUID_BUF, UUID_LEN)
-    local matched_value
+    local matched_value, matched_expr
     if matched_field then
         matched_value = matched_value_len[0] > 0 and
                         ffi_string(matched_value_buf[0], matched_value_len[0]) or
                         nil
+        matched_expr = matched_expr_len[0] > 0 and
```
chobits: empty-string rhs (e.g. `^= ""`) has len 0, so this returns nil even though it matched.

src/interpreter.rs
chobits: not a blocker

Oyami-Srk: I'm wondering if this duplicate slows down the execution because it involves a lot of cloning. Could we do a micro-benchmark for this?

lib/resty/router/context.lua
STARRY-S: ~~Sure, let me try using string reference instead of clone to avoid unnecessary memory copy costs.~~

src/interpreter.rs
STARRY-S: Let me try optimizing it by string reference instead of clone to reduce unnecessary memory copy costs.
Then l'll paste the result of `cargo bench --bench match_mix` to compare the performance changes.

Oyami-Srk: I'm afraid that doesn't work since that requires a `Predicate` lifetime to be in `Match`... And if you want to use cloned `rhs`, it will introduce self-reference too.

From my understanding, we're only missing `rhs` in `matches` for regex:
```
m.matches.insert(
    self.lhs.var_name.clone(),
    Value::String(reg_cap.get(0).unwrap().as_str().to_string()),
);
m.matched_expressions
    .insert(self.lhs.var_name.clone(), rhs.as_str().to_string());
```
A proper fix might be to introduce a new enum type `MatchedValue` like:
```
enum MatchedValue {
    Equal(Value),
    ...
    Regex { captured: Value, expr: Value }
}
```
(Or just using tuple in hashmap)

src/ffi/context.rs
```diff
@@ -210,12 +219,24 @@ pub unsafe extern "C" fn context_get_result(
...
             }
+
+            if !matched_expr.is_null() {
+                assert!(!matched_expr_len.is_null());
+                if let Some(Value::String(v)) = matched.map(MatchedValue::expression) {
+                    *matched_expr = v.as_bytes().as_ptr();
+                    *matched_expr_len = v.len();
+                } else {
+                    *matched_expr_len = 0;
+                }
```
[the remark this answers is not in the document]
STARRY-S: DONE

src/ffi/context.rs
```diff
@@ -210,12 +219,24 @@ pub unsafe extern "C" fn context_get_result(
...
             assert!(!matched_value_len.is_null());
-            if let Some(Value::String(v)) = res.matches.get(matched_field) {
+            let matched = res.matches.get(matched_field);
+
+            if let Some(Value::String(v)) = matched.map(MatchedValue::value) {
                 *matched_value = v.as_bytes().as_ptr();
                 *matched_value_len = v.len();
             } else {
                 *matched_value_len = 0;
             }
```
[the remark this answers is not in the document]
STARRY-S: DONE

lib/resty/router/context.lua
STARRY-S: Done

src/interpreter.rs
STARRY-S: Updated

src/ffi/context.rs
```diff
@@ -210,12 +219,26 @@ pub unsafe extern "C" fn context_get_result(
                 .unwrap();
             assert!(!matched_value.is_null());
             assert!(!matched_value_len.is_null());
-            if let Some(Value::String(v)) = res.matches.get(matched_field) {
+            let matched = res.matches.get(matched_field);
+
+            if let Some(Value::String(v)) = matched.map(MatchedValue::value) {
```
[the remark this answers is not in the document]
STARRY-S: Ignore: panic on non-standard UTF-8 string is expected.

src/ast.rs
```diff
@@ -128,6 +128,34 @@ impl From<String> for Value {
...
+#[derive(Debug)]
+pub struct RegexMatchedValue {
+    pub captured: Value,
+    pub expr: Value,
+}
+
+#[derive(Debug)]
+pub enum MatchedValue {
+    Plain(Value),
+    Regex(Box<RegexMatchedValue>),
```
Oyami-Srk, line 140: Any reason for using `Box` here? `Value` is only 40 bytes; introducing indirection may cause a cache miss.

Oyami-Srk (commented):
Overall LGTM. One question.

src/ast.rs
STARRY-S: The `Value` enum is 40 bytes, and with the `String` key, each hashmap entry is 64 (24 + 40) bytes.

Without `Box`, the `Regex { captured, expr }` makes the enum 80 bytes, so each hashmap entry becomes 104 bytes, which exceeds the 64-byte cache line. This will cause more cache misses, and every `insert/drop` operation on 104 bytes instead of 64 bytes will introduce more `memcpy/memset` operations.

The `Box` heap allocation will only happens on actual regex matches, which is already dominated by the regex engine cost itself.

Quick verification example:

<details>

```rust
use atc_router::ast::Value;
use std::mem::size_of;

enum MatchedValueNoBox {
    Plain(Value),
    Regex { captured: Value, expr: Value },
}

struct RegexMatchedValue {
    captured: Value,
    expr: Value,
}

enum MatchedValueBoxed {
    Plain(Value),
    Regex(Box<RegexMatchedValue>),
}

fn main() {
    println!("Value: {} bytes", size_of::<Value>()); // 40 bytes
    println!("String: {} bytes", size_of::<String>()); // 24 bytes
    println!("Box<RegexMatchedValue>: {} bytes", size_of::<Box<RegexMatchedValue>>()); // 8 bytes
    println!("MatchedValue (no Box): {} bytes", size_of::<MatchedValueNoBox>()); // 80 bytes
    println!("MatchedValue (with Box): {} bytes", size_of::<MatchedValueBoxed>()); // 40 bytes
}
```

</details>

Oyami-Srk: The cache block is usually 64 bytes, but the entire cache line is much larger. I'd assume Kong3 performs more regex operations in the real world because we use regex for traditional compatible mode, IIRC.
Anyway, I don't have a very strong opinion on this; each case has its own trade-off. Thanks for reply!

Oyami-Srk (approved):
LGTM. In theory, it's a breaking change so we're supposed to bump the version to 2.0.0, but in practice I think bumpping minor version should be fine.

chobits (approved):

chobits:
logic doesn't break, ABI breaks. Transparent to pure Lua users; consumers (e.g. kong-ee) using the C API / Rust crate directly need to upgrade.

PR #348, opened 2026-07-16
fix: isolate match state between logical branches

quake:
### Impact scenario

A matcher can capture a request value in one logical branch, fail later in that branch, and then succeed through another branch. The failed branch currently leaves captures and matched values in the final result. Consumers such as Kong can then use stale path captures for URI rewriting or expose them to plugins even though that branch did not determine the match.

### Fix

- evaluate `AND`, `OR`, and `NOT` against branch-local match state
- commit state only from successful positive branches
- preserve router-level reset between separate matcher candidates
- add regressions for nested, negated, alternate, and cross-matcher paths

### Testing

- `cargo fmt --check`
- `cargo test --all-features`
- `36 passed / 0 failed`


StarlightIbuki (approved):
Good catch!
IMO the match object should be returned instead of passed as a parameter. But let's keep the original API

PR #351, opened 2026-08-13
fix(ci): commit Cargo.lock to pin a consistent dependency graph

StarlightIbuki:
## Summary

- Without a checked-in `Cargo.lock`, every CI run resolves dependencies fresh against crates.io. `serde_derive 1.0.229` (pulled in transitively via the `criterion`/`bstr` dev/normal dependencies) recently started requiring `syn ^3`, while `pest_derive` (a direct dependency) still needs `syn ^2` — so a fresh resolve now always contains two major versions of `syn`, tripping clippy's `multiple_crate_versions` lint (promoted to a hard error by `deny(warnings)` in `src/lib.rs`). This broke the [Rust Clippy Report](https://github.com/Kong/atc-router/runs/94371856315) check on #341.
- Pin `serde`/`serde_core`/`serde_derive` to `1.0.228` (the last release before the `syn ^3` requirement) so the graph resolves to a single `syn` version, and commit the resulting `Cargo.lock` so CI stops floating to whatever crates.io has at the moment a run happens.
- Add a weekly scheduled workflow (`dependency-drift.yml`) that runs `cargo update` against a throwaway lockfile and then build/test/clippy, so this kind of drift is caught proactively instead of surfacing on an unrelated PR. Committing `Cargo.lock` only pins *this repo's own* CI builds — it has no effect on what versions downstream consumers of the library resolve to, since Cargo never uses a dependency's own lockfile — so the periodic job is there to keep that pinned graph from going stale.
- Closes in favor of / supersedes #350, which took the alternative approach of just allow-listing the lint. Pinning the lockfile is more robust since it also prevents *other* transitive crates from silently drifting into similar conflicts (or breaking builds outright) between CI runs.

## Test plan

- [x] `cargo generate-lockfile` + pin `serde`/`serde_core`/`serde_derive` to `1.0.228` → single `syn 2.0.119` entry in `Cargo.lock`
- [x] `cargo clippy --manifest-path ./Cargo.toml --locked -- -W clippy::correctness -W clippy::cargo -W clippy::suspicious -W clippy::style` locally: no `multiple_crate_versions` error (only the two pre-existing, unrelated `needless_borrow` style warnings remain)
- [x] `cargo test --workspace --locked`: all tests pass
- [x] Simulated the new weekly job locally: `cargo update` (undoing the pin) + same clippy invocation reproduces the `multiple_crate_versions` error cleanly, confirming the job would catch this exact class of drift
- [ ] CI passes on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/dependency-drift.yml
```diff
@@ -0,0 +1,42 @@
...
+      run: |
+        cargo test --workspace --all-targets --locked
+        cargo test --workspace --doc --locked
+
+    # Mirrors the invocation in lint.yml's rust-lint action, so this only
+    # flags regressions caused by fresher dependencies, not lints outside
+    # what the PR-time check already covers. --locked here just pins to
+    # what the 'cargo update' step above already resolved.
+    - name: 'Run cargo clippy'
+      run: cargo clippy --manifest-path ./Cargo.toml --locked -- -W clippy::correctness -W clippy::cargo -W clippy::suspicious -W clippy::style
```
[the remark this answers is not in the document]
windmgc: @StarlightIbuki  ^

windmgc (approved):

.github/workflows/dependency-drift.yml
Oyami-Srk, line 1: We should remove this workflow before https://github.com/rust-lang/cargo/pull/17335 is stabilized. This workflow may put us in a dangerous place where a supply chain attack might happen.

For example, recently, a supply chain attack on `arrayref` (you can check `#security-incident-arrayref-typosquat-houdini-compromise` Slack channel). If the same happened and this workflow was executed within the time window, we'd be doomed.

StarlightIbuki: Good call. Turning this into a draft

windmgc: @Oyami-Srk Isn't this protected by the `--locked` param which has already existed in this PR?

Oyami-Srk: This workflow calls `cargo update`, which will resolve to the latest version and update the lockfile.

windmgc: Ah yes, you're right

Oyami-Srk: Update: it will be stable in Rust 1.100.0, possibly released on 12 November, 2026.

